PR #3, opened 2020-03-15
Allow users to sign up for being notified

florianschmidt1994:
This PR includes functionality for users to sign up when
no help request is currently available in their region.
This is still somewhat WIP as I still want to add functionality
to always display this field but I would appreciate it if you
guys could maybe test the happy flow and report any obvious oversights!


src/components/FilteredList.js
```diff
@@ -1,11 +1,65 @@
...
+
+  const [email, setEmail] = useState('');
+  const [signInLinkSent, setSignInLinkSent] = useState(false);
+
+  const handleClick = async () => {
+    window.localStorage.setItem('emailForSignIn', email);
+
+    try {
+      await firebaseApp.auth().sendSignInLinkToEmail(email, {
+        url: 'http://localhost:3000/#/complete-offer-help?location=' + location,
```
thepiwo: this url should probably be not linked to the environment or production

src/components/FilteredList.js
```diff
@@ -1,11 +1,65 @@
...
+  const {location} = props;
+
+  const [email, setEmail] = useState('');
+  const [signInLinkSent, setSignInLinkSent] = useState(false);
+
+  const handleClick = async () => {
+    window.localStorage.setItem('emailForSignIn', email);
+
+    try {
+      await firebaseApp.auth().sendSignInLinkToEmail(email, {
```
thepiwo: maybe just await `.catch` to handle errors instead of try catch

src/components/FilteredList.js
```diff
@@ -1,11 +1,65 @@
...
-export default function FilteredList () {
+const NotifyMe = (props) => {
+
+  const {location} = props;
+
+  const [email, setEmail] = useState('');
+  const [signInLinkSent, setSignInLinkSent] = useState(false);
+
+  const handleClick = async () => {
+    window.localStorage.setItem('emailForSignIn', email);
```
thepiwo: can we solve this in a more react way? (I don't know how)

florianschmidt1994: This is the recommended way by firebase

PR #14, opened 2020-03-16
Redesign landing page warnings

florianschmidt1994:


kenodressel:
the url in the pdf is wrong, it should be quarantaenehelden.org with the ä

PR #42, opened 2020-03-18
Making zip code and email required in /notify-me

### LINKED ISSUE #35: Make PLZ field required in /notify-me
We entries where the plz is `null` and we got a user request that he accidentally submitted an empty form. This should be handled in the frontend.

Maurice22:
Disables notify button until zip code and email are non-empty

Closes #35


Maurice22:
Let me check

Maurice22:
Now based on HTML5 forms and required attribute. What do you think?

(BTW, this also enabled email validation)

PR #44, opened 2020-03-18
show correct info text if no filtered entries were found

### LINKED ISSUE #16: Show error on no result
The FilteredList.js component should show an error if the search did not yield any results to inform the user that the search is completed.

jagaSto:
* fixes #16 --> currently, the info text is not shown if the search returned no results
* updated info text (removed redundant word)

Screenshot 2020-03-18 at 21 37 16

@florianschmidt1994


florianschmidt1994:
Thanks for contributing and for adding the screenshot right away :) LGTM, merging

Issue #18, opened 2020-03-17
Add Pagination to frontpage


kenodressel:
Cloud Firestore has several features to help you manage queries that return a large number of results:

- Cursors, which allow you to resume a long-running query.
- Page tokens, which help you paginate the query results.
- Limits, which specify how many results to retrieve.
- Offsets, which allow you to skip a fixed number of documents. (DO NOT USE, 1 read per skipped doc)

Maurice22:
I am working on this over at User/Repository@SHA: Maurice22/quarantine-hero@8c87da33746d67563e31d7301e4606f0f6bb3d1e

Maurice22:
So there is an issue: If we use pagination, client-side filtering of locations does not work anymore (as we won't search the whole database). I wanted to fix this by filtering the location on the database as well but it turns out that more or less firebase only allows for one filter.

Because of that I would suggest to disable pagination if a user searches for a location. What do you think? Do you have other ideas?

kenodressel:
@Maurice22 Yes thats a problem but it also means we significantly reduce reads.
Our plan for today is to re-enable the google maps integration and therefore also re-enable a proper search. So I think its fine if you build it and we merge it after we re-enable the maps api.

Maurice22:
@kenodressel Please check out my version at Maurice22/quarantine-hero@383f2c7246739be42cff0dc542c75cfca004a6ba . I completely redesigned the way queries are done, including support when maps API is enabled. Can you give it a test? If everything works fine, I will create a PR for it.

Maurice22:
Ah yeah and of course I added pagination there :D Check the commit messages for more info.

kenodressel:
I took a look and it seems fine to me. Are you available for a call (send mail to help@quarantaenehelden.org) so we can include you better in our planning?

Maurice22:
Closed in #46

PR #46, opened 2020-03-19
Pagination Feature

Maurice22:
Implements #18 

Also redesigns the way requests are filtered by filtering them server-side.

@kenodressel Might this be counterproductive in terms of readcount?


Maurice22:
@kenodressel Now on the most recent version. I also added a "ScrollToTop"-Button component from npm that is visible on all sites as soon as you start scrolling down.

src/components/FilteredList.js
```diff
@@ -7,62 +7,139 @@ import LocationInput from './LocationInput';
...
-  const [entries, setEntries] = useState([
-    {
-      id: 'placeholder-id',
-    }]);
-  const [filteredEntries, setFilteredEntries] = useState([
-    {
-      id: 'placeholder-id',
-    }]);
+  const [entries, setEntries] = useState([]);
+  const [scheduledSearch, scheduleSearch] = useState([]);
```
kenodressel: should be `setScheduledSearch` to follow conventions

src/components/FilteredList.js
```diff
@@ -7,62 +7,139 @@ import LocationInput from './LocationInput';
...
+        queryResult = collection.orderBy('d.timestamp', 'desc');
+        console.error('Error', error);
+      }
+    } else {
+      queryResult = collection;
+
+      if (location && location !== '') {
+        queryResult = queryResult.orderBy('d.plz', 'asc');
+
+        var lowerLimit = Number.parseInt(location) * (Math.pow(10, Math.max(5 - location.length, 0)));
```
kenodressel: this looks to me like we can only search for postal codes with 5 digits, but we should be also able to find 4 digit ones

src/components/FilteredList.js
```diff
@@ -7,62 +7,139 @@ import LocationInput from './LocationInput';
...
+      }
+      scheduleSearch(setTimeout(() => {
+        loadFilteredData(buildQuery(address));
+      }, 500));
     }
   };
 
   const NoHelpNeeded = (props) => {
-    return <div className="w-full text-center my-10">In {location} wird gerade keine Hilfe gebraucht!</div>
+    return <div className="w-full text-center my-10">In {location} wird gerade aktuell keine Hilfe gebraucht!</div>
```
kenodressel: lets also add `font-open-sans` here

Maurice22:
@kenodressel  Changes incorporated

PR #68, opened 2020-03-20
Adding dynamic "More content" button

### LINKED ISSUE #53: Establish an arrow down button on the main page
Right now, the main page does not show the "Help"-Buttons. We should create an overlay "arrow" (similar to the one further down) that is tied to the screen bottom (unless one scrolls down) that visualizes that there is more content following.

Maurice22:
fixes #53


kenodressel:
Two things:
- I am not happy with it design wise since it just covers some text and there is now shadow or background element to cover it
- The headline "Aktuelle Anfragen" needs a bit more spacing towards the top

Maurice22:
@kenodressel New version available, please review

PR #78, opened 2020-03-20
Fix temporary redirect on dashboard page after reload when logged in

timjb:
This was due to not waiting for authentication from firebase before
redirecting to the signup page. Now, the value `undefined` is used when
no authentication information is available and the special string
'NOT_LOGGED_IN' is used to denote the state where it has been determined
that the user is not logged in.

I have tested that

* still redirects to the signup page when not logged in and
* the dashboard page still shows notifications.

I have not tested that the dashboard page still shows requests for help and their responses because I didn't want to create a fake request for help just for testing purposes.

### DIFF
--- a/src/views/Dashboard.jsx
+++ b/src/views/Dashboard.jsx
@@ -3,48 +3,47 @@ import { Redirect, Link } from 'react-router-dom';
 import fb from '../firebase';
 import Entry from '../components/Entry';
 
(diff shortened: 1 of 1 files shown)


florianschmidt1994:
This looks good, thanks for contributing! I've also tested the functionality and it still works for me as well. 

We've recently changed our branching model to work on `development` and only create releases on `master` periodically. @kenodressel are you d'accord if I still merge this into master?

PR #91, opened 2020-03-20
adds new logos and articles

kenodressel:


Maurice22:
Updated with own changes, passing forward to @florianschmidt1994 for review

florianschmidt1994:
image
I don't know what the design is on this, but could we maybe add some space between the last line "Pressekontakt" and the one before? Also maybe make the icon vertical center?

Maurice22:
This should not look like this tbh. have you pulled the most recent version?

florianschmidt1994:
Now I do :) LGTM

PR #92, opened 2020-03-20
Feature: Add reporting feature for malicious posts

### LINKED ISSUE #27: Add reporting feature for malicious posts

jagaSto:

* fixes #27 

**Checklist**
- [x] `yarn build` passes

## Preview (if logged in)

* find the preview [here](https://quarantaenehelden.slack.com/archives/CV29P59A6/p1584902920049500), two cases showcased:
* case 1: existing user is not logged in, gets redirected after logging in
* case 2: new user tries to report an entry, gets redirected after signing up

## To discuss

* reporting state is not persisted once the posting is selected and redirected to `/offer-help/:postingId` --> fine or should the state be persisted somehow? if yes, any suggestions on how to do so?
* ~if we want to store the user id of the user who reported the post, we need to make sure that the functionality is only available for logged-in users. The alternative would be to always show the button and not store the userId in the collection. Ideas?~--> redirect user to `/signup`, as discussed

## TODOS before merge

* ~make sure firebase function is set up correctly :)~

## Screenshots

Screenshot 2020-03-22 at 19 55 49

Screenshot 2020-03-22 at 19 55 54

Screenshot 2020-03-22 at 19 42 40


jagaSto:
@kenodressel

kenodressel:
I think only the minority of people are logged in so I would think about just adding the button without forcing the login or showing in anyways but redirecting users to login before they can report. Opinions? @florianschmidt1994 @Maurice22

Maurice22:
Seems reasonable to me! Will also potentially reduce abuse.

jagaSto:
Sounds good! So redirect to `/signup` if not logged in. Should we show a different message once the user arrives on the `signup` page, e.g., to tell him that he needs to log in before being able to report a post? Once he is signed in, should it redirect to the original post that he wanted to report?

kenodressel:
The redirect sounds good after the report sounds good. You might need to adjust the signup component :)

jagaSto:
@kenodressel updated! See the description of the PR for the link to the demo. As discussed, if a user tries to report a post but is not logged in, he is redirected to the signup page. After he signs up/logs in, he is redirected to the `/offer-help/:id` page for the entry he tried to reported before. I tested both cases for an existing user that is not logged in, and for a new user that does not have an account yet. I also updated the signup screen to show a different reason for signing up if we detect that the user comes from that previous route! Let me know what you think :)

src/views/Signup.jsx
```diff
@@ -13,6 +18,9 @@ const Signup = (props) => {
   const [password, setPassword] = React.useState('');
   const [error, setError] = React.useState('');
   const [passwordResetSuccess, setPasswordResetSuccess] = React.useState(false);
+  const history = useHistory();
+  const location = useLocation();
+  const userPreviouslyAttemptedToReportAnEntryButWasNotLoggedIn = location && location.state && location.state.reason === 'report_entry_user_not_logged_in';
```
kenodressel: could we pick a shorter variable name? :D

src/views/Signup.jsx
```diff
@@ -37,6 +45,11 @@ const Signup = (props) => {
       if (!signInResult.user.emailVerified) await signInResult.user.sendEmailVerification();
     }
     if (signUpResult.code) setError(signUpResult.message);
+
+    // redirect to entry the user wanted to report
+    if (userPreviouslyAttemptedToReportAnEntryButWasNotLoggedIn && location.state && location.state.entry_id) {
```
kenodressel: could you refactor the returnUrl above so we dont have two redirect mechanisms? I like yours better with using the state

src/views/Signup.jsx
```diff
@@ -51,13 +64,17 @@ const Signup = (props) => {
       .catch(() => setError('Fehler beim Passwort zurücksetzen. Bist du sicher, dass es seine E-Mail ist?'));
   };
 
+  const reasonForRegistration = userPreviouslyAttemptedToReportAnEntryButWasNotLoggedIn
+    ? 'um den Beitrag zu melden'
+    : 'um eine Hilfe-Anfrage zu posten';
+  const textToShow = `Registriere dich mit deiner E-Mail und einem Passwort ${reasonForRegistration} oder melde dich an, wenn du bereits einen Account besitzt.`;
+
```
Maurice22: Ca we pass this as an argument and do the redirecting generic as before? This would keep the code cleaner and better expandable in the future.

Maurice22:
With regards to persistency, shouldn't we save the "reporters" in the ask-for-help collection? Then it will be loaded everytime the entry is loaded. If the reporterIds array includes the id of the authenticated user, we would show "Gemeldet".

src/views/Signup.jsx
Maurice22: You should be able to use the Elvis operator here. i.e. `location?.state?.reason === `

jagaSto: Looks like a cool feature, but is it already available without babel? 🤔 https://stackoverflow.com/a/54482814/8547462

jagaSto: > With regards to persistency, shouldn't we save the "reporters" in the ask-for-help collection? Then it will be loaded everytime the entry is loaded. If the reporterIds array includes the id of the authenticated user, we would show "Gemeldet".

I was also thinking about this initially, but I think @kenodressel argued to put it in its own collection initially since we won't act on it in the beginning anyway. Happy to revisit thoguh!

jagaSto:
> With regards to persistency, shouldn't we save the "reporters" in the ask-for-help collection? Then it will be loaded everytime the entry is loaded. If the reporterIds array includes the id of the authenticated user, we would show "Gemeldet".

I was also thinking about this initially, but I think @kenodressel argued to put it in its own collection initially since we won't act on it in the beginning anyway. Happy to revisit thoguh!

src/views/Signup.jsx
jagaSto: sure thing

src/components/Entry.jsx
```diff
@@ -38,6 +40,30 @@ export default function Entry(props) {
...
+    e.preventDefault();
+
+    const collectionName = 'reported-posts';
+    const reportedPostsCollection = fb.store.collection(collectionName);
+
+    // redirect the user to the login page, as we can only store user ids for logged-in users
+    if (!fb.auth.currentUser || !fb.auth.currentUser.uid) {
+      const pathToOfferHelp = `offer-help/${id}`;
+      const pathname = `/signup/${encodeURIComponent(pathToOfferHelp)}`;
+      return history.push({ pathname, state: { reason_for_registration: 'den Beitrag zu melden' } });
```
jagaSto: simplified this together with @Maurice22 to make the redirect behavior in the signup component more consistent (any redirect is executing using the [`returnUrl `](https://github.com/kenodressel/quarantine-hero/pull/92/files#diff-76aca45a1137d7d9ef9ca65ab00c6ee7R30))

src/components/Entry.jsx
```diff
@@ -79,6 +131,29 @@ export default function Entry(props) {
         {' '}
         {t('components.entry.needsHelp')}!
       </span>
+
+      {(attemptingToReport === false)
+        ? (
+          <button type="button" className={initializeReportEntryButtonClass} onClick={initializeReportEntry}>
+            {/* eslint-disable-next-line import/no-dynamic-require */}
```
jagaSto: Screenshot 2020-03-23 at 11 06 52


🤷‍♂

kenodressel: ist erstmal nicht so wichtig

src/views/OfferHelp.jsx
```diff
@@ -63,7 +63,7 @@ export default function OfferHelp() {
         <div className="mt-4 p-1 font-teaser">
           {t('views.offerHelp.replyToRequest')}
         </div>
-        <Entry {...entry} showFullText highlightLeft />
+        <Entry {...entry} showFullText highlightLeft key={entry.id} />
```
Maurice22: Is this even needed? I can't find any reference in the Entry component

kenodressel: This is done to prevent react warnings if a component is used twice. So I think we can leave it

jagaSto: yes this is necessary since the component is rendered multiple times, and only completely created newly if a key attribute is set. If the component is only rerendered, the state is not wiped correctly, which causes issue with the currently implementation. We also do it like this in other places:
* https://github.com/kenodressel/quarantine-hero/blob/e8cfe3ae86eaeb741dbc405b622c5e3bf333c159/src/views/Dashboard.jsx#L84 
* https://github.com/kenodressel/quarantine-hero/blob/e8cfe3ae86eaeb741dbc405b622c5e3bf333c159/src/components/FilteredList.jsx#L171

resource: https://reactjs.org/docs/reconciliation.html#keys

Maurice22: 👌

PR #93, opened 2020-03-20
Internationalization

yorickreum:
As discussed in https://github.com/kenodressel/quarantine-hero/issues/24.
This pull request ist still work in progress. I am happy to continue working on this anytime.
Since it is a huge amount of (trivial) work and affects almost every file, we/you need to decide if you want me to continue this path?
Maybe we should even discuss to merge this pull request in steps since it does not cause any harm if not completed and every change of you at one of the files leads to merging…

PR #94, opened 2020-03-20
Added basic setup for internationalisation

yorickreum:
As decided in https://github.com/kenodressel/quarantine-hero/pull/93, here you find only the basic setup for using i18n-react. I also added a really short example of how to use it in App.jsx.


kenodressel:
Could we remove the english version. I have an english OS so I get a very partial translation. I would say we start with German only. Other than this the PR looks fine.

yorickreum:
Done. I left the file (for the future?) but deleted all of its content.

PR #96, opened 2020-03-21
Make base URL configurable for local dev

timjb:
This makes it possible to test authentication locally when developping on
localhost:3000 by replacing `src/appConfig.jsx` with
`src/appConfig.dev.jsx`.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+REACT_APP_BASE_URL=https://www.quarantaenehelden.org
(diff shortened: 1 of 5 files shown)


kenodressel:
Could we use ENV variables to automatically switch between the environments?

timjb:
@kenodressel Good idea, I will try this out!

timjb:
@kenodressel It worked, please have another look at this PR :-)

We could use the same mechanism to make it more easier to

* use another firebase configuration and to
* set feature flags

during development. If you agree, I would start working on that after this PR is merged.

kenodressel:
Yes I think its a good idea but we should be careful to always include all env varibales in the repo.

PR #98, opened 2020-03-21
Bug fix: add local sorting if location filter is active

### LINKED ISSUE #89: Timestamp sorting not working while searching
Filtered list should be sorted client-site if search is active.

Background: We can only sort by timestamp if we don't search in firebase. Hence, if the search flag is set, we need to do sorting locally.

jagaSto:
* fixes #89 

documentation for helper function: https://lodash.com/docs/4.17.15#orderBy

## Before

77228845-64431c00-6b8a-11ea-96cb-faa2336331a3

## After

Screenshot 2020-03-21 at 16 12 17

### DIFF
--- a/src/components/FilteredList.jsx
+++ b/src/components/FilteredList.jsx
@@ -99,12 +99,12 @@ export default function FilteredList(props) {
     });
   };
 
-  const loadFilteredData = async (queryPromise, address) => {
+  const loadFilteredData = async (queryPromise) => {
     const query = await queryPromise;
     const value = await query.get();
 
     // no location filter applied
-    if (!address) {
+    if (!location) {
       appendDocuments(value.docs);
       return;
     }
@@ -122,7 +122,7 @@ export default function FilteredList(props) {
         clearTimeout(scheduledSearch);
       }
       setScheduledSearch(setTimeout(() => {
-        loadFilteredData(buildQuery(address), address);
+        loadFilteredData(buildQuery(address));
       }, 500));
     }
   };
@@ -134,7 +134,7 @@ export default function FilteredList(props) {
         clearTimeout(scheduledSearch);
       }
       setScheduledSearch(setTimeout(() => {
-        loadFilteredData(buildQuery(address), address);
+        loadFilteredData(buildQuery(address));
       }, 500));
     }
   };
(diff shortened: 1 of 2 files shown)


kenodressel:
Could we solve this natively without an additional dependency?  For example https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/sort could work. An example implementation of this could be found in https://github.com/florianschmidt1994/quarantaenehelden-firebase-functions/blob/master/functions/index.js#L103

Maurice22:
+1 for doing it with array sort

src/components/FilteredList.jsx
```diff
@@ -99,11 +100,20 @@ export default function FilteredList(props) {
     });
   };
 
-  const loadFilteredData = async (queryPromise) => {
+  const loadFilteredData = async (queryPromise, address) => {
```
Maurice22: Instead of introducing a new variable here, we can use the global `searching` variable

Maurice22: Instead of having an additional variable here, we should use the global `searching` variable

Maurice22: Although this might not be updated at this point.. I vote for rewriting the whole logic to make it leaner.

jagaSto: Hmm, I don't know about that, `searching` only seems to be set to true if the firebase query is executed.

jagaSto:
@kenodressel @Maurice22 updated to use native `Array.sort()`. Please let me know if I should further refactor the logic, as requested by @Maurice22 or if we can leave it as-is for now!

jagaSto:
Just discovered `location` is available, will refactor!

src/components/FilteredList.jsx
Maurice22: Yes, this is because it is set by react and this is unfortunately not done synchronously. Will review it like this for now, but we should create an issue to refactor the whole code here. Maybe once we reenabled the maps api to make it less cluttered.

jagaSto:
done

src/components/FilteredList.jsx
jagaSto: Just discovered that the `location` variable is available from state --> refactored :)

Maurice22:
The issue with those variables is that they are set asynchronously by react. I need to check if location will always be available when we need it. I think I will refactor the code in this branch before continuing with the review.

jagaSto:
I'm pretty sure it is always available (assumption based on my testing), but feel free to test yourself :)

kenodressel:
I think `location` should always be there. At least with the default value (empty string). So I am fine with this state.

Maurice22:
Alright, I added issue #100 for future refactoring.

PR #113, opened 2020-03-22
Fix usages of addEventListener in useEffect hook

timjb:
These hooks should return a function that removes the event listener.
Otherwise, there could be a reference to a component that has already
been removed, which causes an error to be logged in the console during
development.

There should be no user-visible changes (except perhaps slightly better performance / lower memory usage).

**Checklist**
- [x] `yarn build` passes
- [x] `yarn lint` does not introduce any new errors (We know there are currently 12 errors and 9 warnings and working hard on removing them!)

**Others details**
- [ ] This PR introduces a new dependency.
  
If you have any questions be sure to let us know! For review ping anyone in this list:
 
 - @florianschmidt1994
 - @kenodressel
 - @Maurice22
 - @tgraupne

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -30,6 +30,7 @@ import NotifyMe from './views/NotifyMe';
 import ScrollToTop from './components/ScrollToTop';
 import ShareButtons from './components/ShareButtons';
 import Presse from './views/Presse';
+import createEventListener from './util/createEventListener';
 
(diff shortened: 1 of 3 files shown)


florianschmidt1994:
Thanks for fixing this! Didn't it worked this way, I've learned something new today, yay 👍 LGTM

PR #120, opened 2020-03-22
Disable single use mail

### LINKED ISSUE #36: Don't allow single-use emails on signup
Blacklist some well-known "throw-away-email" domain for user registration.

Reference:
https://github.com/johnduhart/throwaway-email-blacklist/blob/master/domains.txt
https://www.internet-neukunden.de/2009/02/22/liste-mit-57-anbieter-von-wegwerf-emails-antispam/
https://videonerd.de/wegwerf-e-mail-adressen/

funtonia:
fixes #36 

Whenever users have to enter their mail address (on NotifyMe, SignUp, OfferHelp), it is checked whether the used one is a single-use-address by comparing it to a static list of known single-use-domains. The HTML5 form validation is used to show a specific error message whenever a single-use-address is used, disabling the sign up with it.


src/views/NotifyMe.js
```diff
@@ -34,6 +33,21 @@ export default function NotifyMe() {
...
+      if (dems.includes(domain)) {
+        // this is a DEM
+        data.target.setCustomValidity("Bitte keine Wegwerf-Emailadresse verwenden.");
+      } else {
+        // this is not a DEM
+        data.target.setCustomValidity("");
+      }
+    }
+  };
+
```
tinobo: Is there a way to make this code reusable to reduce code duplication?

funtonia: Sure! On it :)

funtonia: Done with commit 8ad5a56675e3c0b44148e4d2637d9fb126cae447.

Maurice22:
@funtonia, looks like your branch is based on a pretty old version of the code. Please merge the current develop branch, resolve conflicts and check for linting errors, thanks

src/components/MailInput.js
```diff
@@ -0,0 +1,25 @@
...
+    var domain = mail.substring(mail.lastIndexOf("@") +1);
+    // check if the mail address is included in the DEM-list
+    if (dems.includes(domain)) {
+      // this is a DEM
+      data.target.setCustomValidity("Bitte keine Wegwerf-Emailadresse verwenden.");
+    } else {
+      // this is not a DEM
+      data.target.setCustomValidity("");
+      props.onSetEmail(mail);
+    }
```
Maurice22: You should reset the custom error message in case the email does not include an @. Otherwise, when you entered a DEM once and delete the address again, the custom error stays.

src/components/MailInput.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import dems from '../assets/dems.json';
+
+function checkForDEM(data, props) {
+  const mail = data.target.value;
+  if(mail.includes('@')) {
```
timjb: `String.prototype.includes()` is [not supported by IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Browser_compatibility). As far as I can see, you are not using a polyfill library. Therefore, I believe that this will cause an exception in IE, which will prevent any IE user from sending a request for help. (Can't test it because I am using Linux.)

src/components/MailInput.js
```diff
@@ -0,0 +1,25 @@
...
+    var domain = mail.substring(mail.lastIndexOf("@") +1);
+    // check if the mail address is included in the DEM-list
+    if (dems.includes(domain)) {
```
timjb: `Array.prototype.includes()` is [not supported by IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility). As far as I can see, you are not using a polyfill library. Therefore, I believe that this will cause an exception in IE, which will prevent any IE user from sending a request for help. (Can't test it because I am using Linux.)

funtonia: Thanks for pointing that out, had not heard of this issue before. Changed it to check via indexOf() as you suggested, works like a charm!

funtonia:
closing this and moving to  #125 as merging this was way more complicated than starting from scratch

PR #124, opened 2020-03-23
Feature/google maps search

Maurice22:

Enabling the google maps api search. This PR, sets the maps API feature flag to true, which means new requests, etc. are created using the geocords in addition to the zip-code. Furthermore, we enable the search based on geolocation instead of zip-code. Therefore, not only a new dependency was introduced, but also the way request are loaded was reworked to be leaner.

**Checklist**
- [x] `yarn build` passes

**Others details**
- [ ] This PR introduces a new dependency. Reason: Added react slider to enable selection of a search range.


src/components/FilteredList.jsx
```diff
@@ -17,63 +18,47 @@ export default function FilteredList(props) {
 
   const [searching, setSearching] = useState(false);
   const [location, setLocation] = useState('');
+  const [radius, setRadius] = useState(10);
```
jagaSto: we could rename this to `radiusInKm` or add a comment, to indicate the unit for this property :)

src/views/AskForHelp.jsx
```diff
@@ -51,7 +53,11 @@ export default function AskForHelp() {
   const handleSelect = (address) => {
     setLocation(address);
     geocodeByAddress(address)
-      .then((results) => getLatLng(results[0]))
+      .then((results) => {
```
jagaSto: could we use async/await syntax here? Generally, I find it easier to read than promise chains

public/locales/de/translation.json
```diff
@@ -203,7 +203,7 @@
     },
     "locationInput": {
       "yourPostalCodeOrNeighbourhood": "Deine Postleitzahl oder Nachbarschaft...",
-      "yourPostalCode": "Deine Postleitzahl oder Nachbarschaft..."
+      "yourPostalCode": "Deine Postleitzahl..."
```
jagaSto: 😆

src/components/FilteredList.jsx
```diff
@@ -83,61 +68,73 @@ export default function FilteredList(props) {
...
-      appendDocuments(value.docs);
-    });
-  };
-
-  const loadFilteredData = async (queryPromise) => {
+  const loadDocuments = async (queryPromise, searchActive) => {
     const query = await queryPromise;
-    const value = await query.get();
+    const results = await query.get();
+    setEntries([]);
```
jagaSto: why do we set the entries to an empty array here?

Maurice22: Because we need to clear all entries before we set the new ones, i.e. if someone does a search, the old entries need to be cleared before the new ones are added

src/components/FilteredList.jsx
```diff
@@ -17,63 +18,47 @@ export default function FilteredList(props) {
...
+    if (isMapsApiEnabled) {
       try {
-        const results = await geocodeByAddress(loc);
+        const results = await geocodeByAddress(searchAttr);
         const coordinates = await getLatLng(results[0]);
-        queryResult = queryResult.near({ center: new fb.app.firestore.GeoPoint(coordinates.lat, coordinates.lng), radius: 30 });
+        return geoCollection.near({ center: new fb.app.firestore.GeoPoint(coordinates.lat, coordinates.lng), radius });
       } catch (error) {
-        queryResult = collection.orderBy('d.timestamp', 'desc');
+        // Fallback
```
jagaSto: should we add error logging here? Not sure what our general logging policy is so far

src/components/FilteredList.jsx
```diff
@@ -83,61 +68,73 @@ export default function FilteredList(props) {
...
     if (!isMapsApiEnabled) {
+      // Stop any scheduled search task
       if (scheduledSearch) {
         clearTimeout(scheduledSearch);
       }
-      setScheduledSearch(setTimeout(() => {
-        loadFilteredData(buildQuery(address));
-      }, 500));
+
+      // If address is non-empty, schedule new search task to only start searching after user stopped typing
```
jagaSto: nice!

PR #135, opened 2020-03-23
Refactor Dashboard component

timjb:

* use useCollectionDataOnce hook from react-firebase-hooks library for
  easier loading of data
* separate authentication into a separate component
* use placeholder when authentication information or data is loading
* pull definition of Notification function component out of Dashboard
  function component. Nesting function components in this way is bad
  because it prevents React from reusing parts of the virtual DOM,
  since every time the outer component is executed, the inner component
  is redefined.

**Checklist**
- [x] `yarn build` passes

**Others details**
- [ ] This PR introduces a new dependency. Reason: _Mention your reason for introducing this dependency here_
  
If you have any questions be sure to let us know! For review ping anyone in this list:
 
 - @florianschmidt1994

### DIFF
--- a/public/locales/de/translation.json
+++ b/public/locales/de/translation.json
@@ -73,6 +73,7 @@
       "toYourOverview": "ZU DEINER ÜBERSICHT"
     },
     "dashboard": {
+      "isLoading": "Daten werden geladen …",
       "youWillBeNotified": "Du wirst benachrichtigt, wenn jemand in der Nähe von",
       "needsHelp": "Hilfe benötigt",
       "yourRequests": "Deine Hilfegesuche",
--- a/src/views/Dashboard.jsx
+++ b/src/views/Dashboard.jsx
@@ -32,10 +32,11 @@ function Notification(props) {
   );
 }
 
-function DashboardLoading() {
-  const { t } = useTranslation();
-  return <div className="font-open-sans my-8 text-lg text-center">{t('views.dashboard.isLoading')}</div>;
-}
+// Commented out until there is a consistent way of showing placeholders on the site
+// function DashboardLoading() {
+//   const { t } = useTranslation();
+//   return <div className="font-open-sans my-8 text-lg text-center">{t('views.dashboard.isLoading')}</div>;
+// }
 
 function Dashboard(props) {
   const { user } = props;
@@ -57,7 +58,8 @@ function Dashboard(props) {
   const offers = (offersDocs || []).map((val) => ({ ...val.d, id: val.id }));
 
   if (isLoadingRequestsForHelp || isLoadingOffers) {
-    return <DashboardLoading />;
+    // Commented out until there is a consistent way of showing placeholders on the site
+    // return <DashboardLoading />;
   }
 
   return (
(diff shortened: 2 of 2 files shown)


public/locales/de/translation.json
```diff
@@ -75,6 +75,7 @@
       "toYourOverview": "ZU DEINER ÜBERSICHT"
     },
     "dashboard": {
+      "isLoading": "Daten werden geladen …",
```
jagaSto: Should we move the `...` into the component? Then the property only contains the translation, and we won't have to add the dots for every language :)

src/views/Dashboard.jsx
```diff
@@ -1,76 +1,74 @@
...
 
-const getUserData = async (currentUser) => {
-  const query = askForHelpCollection.where('d.uid', '==', currentUser.uid);
-  const value = await query.get();
-  const sortedEntries = value.docs
-    .map((doc) => ({ ...doc.data().d, id: doc.id }))
-    .sort((a, b) => b.timestamp - a.timestamp);
-  return sortedEntries;
-};
+const Notification = (props) => {
```
jagaSto: we could also create this with the function syntax, right? To be consistent with the other function components below

src/views/Dashboard.jsx
```diff
@@ -1,76 +1,74 @@
...
-      getUserData(user).then(setEntries);
-      getOffers(user).then(setOffers);
-    }
-  }, [user]);
+  const { t } = useTranslation();
 
-  if (!isAuthLoading && (!user || !user.email)) {
-    return <Redirect to="/signup/dashboard" />;
-  }
+  const [requestsForHelpUnsorted, isLoadingRequestsForHelp] = useCollectionDataOnce(
```
jagaSto: 👍 for the naming!

src/views/Dashboard.jsx
```diff
@@ -1,76 +1,74 @@
...
+  if (isLoadingRequestsForHelp || isLoadingOffers) {
+    return <DashboardLoading />;
+  }
 
   return (
     <div className="p-4">
       <h1 className="font-teaser py-4 pt-10">{t('views.dashboard.yourRequests')}</h1>
 
-      {entries.length === 0
+      {requestsForHelp.length === 0
```
Maurice22: Is all requests are deleted, the message will not be displayed again. It is only shown upon reload.

src/views/Dashboard.jsx
timjb: Done.

public/locales/de/translation.json
timjb: I think that there are reasons for and against this:

1. On one hand, including the ellipsis in the translation makes it possible to localize it. Maybe other languages use other typographic means to signify an ongoing process, who knows? Also, some languages require a space before the ellipsis, while others don't. Making the ellipsis part of the localized text solves that problem.
2. On the other hand, making the ellipsis part of the static part makes it possible to replace it by a spinner, for example.

I will change this if you think that 2 outweighs 1. But saving effort does not seem to be a good reason for changing this to me.

src/views/Dashboard.jsx
timjb: Fixed this (although it was broken before) by using the `useCollectionData` hook (which includes real-time updates) instead of `useCollectionDataOnce`. This also made it possible to remove the set to hidden on delete stuff from the `Notification` component.

Maurice22:
To be discussed before merge: Do we want to show the "Data loading..." or render the sit and let the result render as they come in.

kenodressel:
To follow the current style we propose to remove the "Data loading..." for now and rework the whole app to indicate data loading where necessary in a consistent style.

Maurice22:
I will quickly fix this @timjb  so that you don't have extra effort

PR #153, opened 2020-03-25
Feature/solved issues

jagaSto:

* This PR introduces the `solved issues` feature. Entries with responses can now be marked as solved by the owner, rather than deleting them directly. For this, tabs were introduced to the `dashboard` component, where a user can switch between open and closed entries.
* Additionally, new Popup hints were introduced. For example, if a user tries to delete an open entry, he receives a hint that he can mark it as solved instead. Other popups, such as a general _deletion reassurement_ popup and _a deletion was successful_ popup, were introduced as well.
* fixes button bar on mobile
* skips unit tests for now, as they are flaky. Follow-up ticket: #216 

**Checklist**
- [x] `yarn build` passes

**Others details**
- [x] This PR introduces a new dependency. Reason: 
   * [react-tabs](https://github.com/reactjs/react-tabs) is needed for the tabs list in the `dashboard` component
  * [reactjs-popup](https://github.com/yjose/reactjs-popup) is needed for the popups
  * PR depends on the [this PR](https://github.com/florianschmidt1994/quarantaenehelden-firebase-functions/pull/10) for the updated firebase functions

**Screenshots**

Screenshot 2020-03-27 at 10 31 10

Screenshot 2020-03-27 at 10 35 31

Screenshot 2020-03-27 at 10 34 58

Screenshot 2020-03-27 at 10 35 12

Screenshot 2020-03-27 at 10 35 47


src/components/Entry.jsx
```diff
@@ -50,21 +58,53 @@ export default function Entry(props) {
...
   const handleDelete = async (e) => {
     e.preventDefault();
-    const doc = await fb.store.collection('/ask-for-help').doc(props.id).get();
-    await fb.store.collection('/deleted').add({
-      askForHelpId: doc.id, ...doc.data(),
+    const doc = await fb.store.collection(collectionName).doc(props.id).get();
+    await fb.store.collection('/deleted').doc(props.id).set({
+      collectionName, ...doc.data(),
     });
-    fb.store.collection('/ask-for-help').doc(props.id).delete();
```
jagaSto: this delete is now handled in the corresponding firebase function, which also migrates the responses to the entry

src/components/Entry.jsx
```diff
@@ -95,10 +135,100 @@ export default function Entry(props) {
     numberOfResponsesText = `${responses} ${t('components.entry.repliesReceived')}`;
   }
 
+  const commonButtonClasses = 'px-6 py-3 uppercase font-open-sans font-bold text-center';
+  const positiveActionButtonClasses = `bg-secondary text-white hover:opacity-75 rounded mb-2 block min-w-90 ${commonButtonClasses}`;
+  const deleteButtonClasses = `bg-red-200 text-primary ${commonButtonClasses}`;
+  const invertedDeleteButtonClasses = `text-primary font-medium min-w-90 ${commonButtonClasses.replace('font-bold', '')}`;
+
+  const InitializeDeletionButton = getButtonForPopup(
```
jagaSto: I tried to make this as generic as possible, to avoid redundant code repetitions, since all of the buttons in the popup are similar. Please let me know if you have a better suggestion for how to do this!

src/components/Entry.jsx
```diff
@@ -95,10 +135,100 @@ export default function Entry(props) {
...
+
+  const HintsPopup = () => (
+    <Popup
+      modal
+      open={popupVisible}
+      onClose={(e) => {
+        e.preventDefault();
+        setPopupVisible(false);
+      }}
+      contentStyle={{ width: '30%', padding: '0' }}
```
jagaSto: default styling of the popup expands it over the entire screen and has a padding of `5px` which leaves an annoying white border --> overwrite

src/components/Popup.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import i18n from 'i18next';
+
+export function getPopupContentComponent(heading, firstButtonComponent, secondButtonComponent) {
+  const popupContentClasses = 'p-4 bg-kaki font-open-sans flex flex-col justify-center items-center';
+
+  const textBodyWasYourRequestSuccessful = (
+    <>
+      <p className="mt-2">
+        Lass die Community wissen ob deine Suche erfolgreich war.
```
jagaSto: should those be localized also?

src/views/Dashboard.jsx
```diff
@@ -1,13 +1,17 @@
 import React from 'react';
 import { Redirect, Link } from 'react-router-dom';
 import { useAuthState } from 'react-firebase-hooks/auth';
-import { useCollectionData } from 'react-firebase-hooks/firestore';
+import { useCollectionDataOnce } from 'react-firebase-hooks/firestore';
```
jagaSto: as discussed with @kenodressel , we had to switch back to `useCollectionDataOnce` since the real-time fetching of the collection content was interfering with the popup logic. With `useCollectionData`, the `delete success` popup was only shown until the data was re-fetched, and disappeared as soon as the new data was ready, even if the user did not act on the popup content yet -> bad UX

package.json
```diff
@@ -26,7 +26,9 @@
     "react-scroll-up-button": "^1.6.4",
     "react-share": "^4.1.0",
     "react-slider": "^1.0.3",
+    "react-web-tabs": "^1.0.1",
```
florianschmidt1994: Is there a benefit of using react-web-tabs over the tabs from @material-ui? We already this as a dependency from other parts, such as the side bar

src/components/Entry.jsx
florianschmidt1994: In 'src/styles/index.css' there are some classes that use `apply` in order to combine styles from different classes into one. Maybe we could use something like this as well here?

src/components/Entry.jsx
florianschmidt1994: What is the default of `e` when closing a pop up? Just curios :)

src/components/Entry.jsx
```diff
@@ -95,10 +138,100 @@ export default function Entry(props) {
...
+    >
+      {/* eslint-disable-next-line no-nested-ternary */}
```
florianschmidt1994: Could we find a solution here that does not use nested ternary operators? I find this a little hard to read 😉

src/components/Popup.jsx
```diff
@@ -0,0 +1,52 @@
...
+    <>
+      <p>Bitte beachte, dass es einen Moment dauern kann, bis der Eintrag gelöscht wird.</p>
+      <p>
+        Du kannst natürlich jederzeit eine neue Anfrage erstellen und eine*n Held*in um Hilfe bitten.
+        <i>#strongertogether</i>
+      </p>
+    </>
+  );
+
+  /* eslint-disable-next-line no-nested-ternary */
```
florianschmidt1994: Same as above, maybe can we avoid the nested ternary operator here?

src/components/Popup.jsx
florianschmidt1994: I would say so, if there are no technical reasons why not

src/components/Popup.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import i18n from 'i18next';
+
+export function getPopupContentComponent(heading, firstButtonComponent, secondButtonComponent) {
```
florianschmidt1994: Is there any way we could make a React component out of this instead of a function? I'm thinking maybe something like https://medium.com/@Dane_s/react-js-compound-components-a6e54b5c9992, but I've never tried that myself so not 💯 percent sure

src/services/loadResponses.js
```diff
@@ -0,0 +1,7 @@
+import fb from '../firebase';
```
florianschmidt1994: ❤️  it

florianschmidt1994: We should take this as an example and maybe separate more stuff in the future, looks nice and clean 👍

src/views/Dashboard.jsx
florianschmidt1994: Right now, if you switch the tabs after marking something as completed, it will only show after a refresh, correct? Im not sure if this is related, but we should think about whether there is a better way to solve this

src/components/Entry.jsx
jagaSto: yes, however, it is more than just styling since we also specify a different `onClick` handling and icon for each button..

src/views/Dashboard.jsx
jagaSto: > Right now, if you switch the tabs after marking something as completed, it will only show after a refresh, correct?

yes exactly. The better way would be to use `useCollectionData ` since it updates the collection data automatically, and a user would not have to refresh the page anymore. However, it fucks up the popup logic with the current implementation. Once this is merged, I have a follow-up ticket  to make the automagic work with the popups :)

src/components/Popup.jsx
jagaSto: will do!

src/components/Entry.jsx
jagaSto: of `e` or of the padding?^^ padding default is `5px`

package.json
jagaSto: I started off with material-UI tabs, but they were super painful to customize, so I went with this one since the API turned out to be much better :)

src/components/Entry.jsx
jagaSto: sure thing :) was just following syntax suggestions from other components!

src/components/Popup.jsx
jagaSto: will check it out!

jagaSto:
layout for mobile should be fixed!

Screenshot 2020-03-29 at 13 04 51

jagaSto:
< Are you still working on the tests or do you plan to merge the todos and cover them separately?

@florianschmidt1994 yes, currently finishing up the test cases (didn't not come around to do this in the last two days, and afterwards I want to polish the exact things you mentioned (question mark icon being to big, etc :)

cypress/integration/dashboard.spec.js
```diff
@@ -1,17 +1,23 @@
...
+const dummyUserMail = 'yemac68103@gotkmail.com';
+const dummyUserPw = 'test1234';
 
 context('Dashboard', () => {
 
   beforeEach(() => {
     cy.visit('localhost:3000');
-    loginVerifiedUser('koxife6299@fft-mail.com', 'test1234');
+    // IMPORTANT NOTE: if you use a different account, make sure that it has a verified email, in order to create new postings
+    loginVerifiedUser(dummyUserMail, dummyUserPw);
```
florianschmidt1994, line 14: we have `verified@example.com` with password `test1234` for that purpose

src/styles/App.css
```diff
@@ -44,11 +44,6 @@
     background-color: #87a544;
 }
 
-.font-open-sans {
```
jagaSto, line 47: moved to index.css since its needed there for a composing utility

src/styles/index.css
```diff
@@ -14,6 +20,24 @@
     @apply rounded p-4 bg-secondary text-secondary w-full inline-block text-center font-bold;
 }
 
+.btn-common {
```
jagaSto, line 23: replaces the previously used `commonButtonClasses`

florianschmidt1994:
@kenodressel could you have a look as well?

src/components/entry/Entry.jsx
```diff
@@ -51,21 +61,60 @@ export default function Entry(props) {
...
   const handleDelete = async (e) => {
     e.preventDefault();
-    const collectionName = 'ask-for-help';
     const doc = await fb.store.collection(collectionName).doc(props.id).get();
     await fb.store.collection('/deleted').doc(props.id).set({
       collectionName, ...doc.data(),
     });
     setDeleted(true);
+    setAttemptingToDelete(false);
+    setPopupVisible(true);
```
Maurice22, line 70: Shouldn't this be set to false here?

src/components/entry/Entry.jsx
```diff
@@ -51,21 +61,60 @@ export default function Entry(props) {
...
+  };
+
+  const handleSolved = async (e) => {
+    e.preventDefault();
+    const askForHelpDoc = await fb.store.collection('ask-for-help').doc(props.id).get();
+    const data = askForHelpDoc.data();
+    await fb.store.collection('solved-posts').doc(props.id).set(data);
+    setDeleted(true);
+    setAttemptingToDelete(false);
```
Maurice22, line 79: Why do we set deleted here?

src/components/entry/Entry.jsx
```diff
@@ -96,8 +145,27 @@ export default function Entry(props) {
...
+      setAttemptingToDelete={setAttemptingToDelete}
+      handleSolved={handleSolved}
+      showAsSolved={showAsSolved}
+      handleNewAskForHelp={handleNewAskForHelp}
+      cancelDelete={cancelDelete}
+      handleDelete={handleDelete}
+      backToOverview={backToOverview}
+    />
+  );
+
```
Maurice22, line 165: For me, this looks like a hack.. Shouldn't a generic dialog have some text to pass as well as some button types to pass and register a callback that reds the button type and executes the desired action?

package.json
```diff
@@ -28,6 +28,9 @@
...
+    "react-with-firebase-auth": "^1.3.0",
```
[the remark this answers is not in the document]
jagaSto: yes, merging went wrong. thanks!

src/components/entry/Entry.jsx
```diff
@@ -4,16 +4,22 @@ import formatDistance from 'date-fns/formatDistance';
...
-import ExpandLessIcon from '@material-ui/icons/ExpandLess';
-import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import MailOutlineIcon from '@material-ui/icons/MailOutline';
+import DoneIcon from '@material-ui/icons/Done';
+
 import fb from '../../firebase';
 import Responses from '../Responses';
+import PopupOnEntryAction from '../Popup';
+
+import { ReactComponent as QuestionMarkSvg } from '../../assets/questionmark.svg';
```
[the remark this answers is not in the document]
jagaSto: will do for the other SVGs :)

src/components/entry/Entry.jsx
jagaSto: needs to be set to true to trigger the `deletion confirmation` popup, will add a comment!

src/components/entry/Entry.jsx
jagaSto: we use the `deleted` state to hide/show the entry. Agree that it should be refactored to `visible` instead

jagaSto: Edit: will introduce a second state `solved` instead

tgraupne:
Hey team, what is the current state here?

jagaSto:
@Maurice22 requested changes and suggested to implement them himself, we are currently deciding whether this can be done as a follow up to this PR or if they are release critical and should be done beforehand :)

Maurice22:
I probably won't be able to get this done soon, so I created #228 as my follow-up

Maurice22:
Work on #216 and #228 can be started now.

PR #162, opened 2020-03-26
Fix language for Google Maps API calls to german

### LINKED ISSUE #160: Locations are localized to input
Currently, autocomplete suggestions are localized and are getting used as input for the database. Hence, we could run into issues where the location is not displayed in german (see attachment)
image

Maurice22:

Setting the language for Gmaps API Calls to german to prevent locations being displayed in different languages in the entries.

Fixes #160 

**Checklist**
- [x] `yarn build` passes

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -32,7 +32,7 @@
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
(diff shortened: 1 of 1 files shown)


Maurice22:
Haven't looked into this. For geocoding, we set the region explicitly, so there should be no issues here, but if we include it, the autocomplete might be better tailored towards the DACH region. I would vote for setting it to "DE" then.

PR #164, opened 2020-03-26
Add first set of cypress tests

florianschmidt1994:
This PR adds [cypress](https://www.cypress.io) as a testing framework. This PR will
- add cypress as a dev dependency
- add the basic folder structure
- add some eslint globals which are required in the tests
- add some basic tests for the landing page, which required adding some ids to our CTAs
- add a brief section about testing to our homepage

A follow up will be adding issues for specific user flows and having them covered by tests.

**Checklist**
- [x] `yarn build` passes

**Others details**
- [x] This PR introduces a new dependency. Cypress is our test runner and we need a solid testing framework in order to solve this problem


kenodressel:
How would you execute the tests? If I run `yarn test` no tests are found.

florianschmidt1994:
I didn't add a script to `package.json` yet, but that's a good point. For starters I would say let's alias `yarn test` to `yarn run cypress open` and trigger the tests manually in the cypress UI.

We could have automated testing & CI, but I think we should investigate this at a later point and have the test setup ready for contributers to explore and use

florianschmidt1994:
The new setup (from README)

### Testing
 Tests are written and executed with [cypress](https://www.cypress.io). All tests are stored in the `/cypress` directory. 
 Tests can be run locally in two ways:
 - `yarn test` will run the tests immediately in a headless browser and output the results to stdout
 - `yarn cypress` will open the cypress UI for interactive testing and test debugging

.eslintrc.json
```diff
@@ -9,7 +9,12 @@
     ],
     "globals": {
         "Atomics": "readonly",
-        "SharedArrayBuffer": "readonly"
+        "SharedArrayBuffer": "readonly",
+        "context": true,
+        "describe": true,
+        "it": true,
+        "cy": true,
+        "beforeEach": true
```
tgraupne: Hey @florianschmidt1994, I am no expert, but I guess you added those lines to integrate cypress key words explicitly, right? I just wanted to ask you, if you considered integrating an eslint-cypress-plugin and if so, why you chose to add those rules explicitly?

florianschmidt1994: You're very correct, I've added them for cypress globals. For now I didn't see the necessity to add a plugin, b.c. with this all errors and warnings are gone and I didn't want to add another dependency without vetting it

kenodressel:
@florianschmidt1994 love the ids but could we use `data-cy` attributes to avoid conflicts in the future where we might need the ids? For the recommended best practices check this doc: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

florianschmidt1994:
@kenodressel Do you want the data-cy in the future or already for this PR?

kenodressel:
Already in this PR. Should be a minor change and can be then used as a reference implementation.

florianschmidt1994:
@kenodressel should be good to go now

PR #168, opened 2020-03-26
Feature/separate login form

### LINKED ISSUE #108: Seperate Signup and Login Page / Form
This leads to confusion of users who mistype their emails.

Maurice22:

There are two forms now, one for login and one for sign up. By default, users are redirected to the login form. This form does no longer upsert an account but will only validate the entered credentials. The error messages have been adapted to reflect this behaviour. If a user wants to sign up, he can do so by following a button on the signin page. States are maintained between these forms to ensure custom header texts are shown the same in signin and signup. Also the redirect url is passed to signup.

Fixes #108 

**Checklist**
- [ ] `yarn build` passes


Maurice22:
Now also with internationalization :)

public/locales/de/translation.json
```diff
@@ -154,6 +154,36 @@
     "successOffer": {
       "thankYou": "Vielen Dank, wir haben Deine Nachricht weitergeleitet.",
       "toHome": "ZUR STARTSEITE"
+    },
+    "signIn": {
+      "noUser": "Es existiert kein Nutzer mit dieser Email. Bitte registriere dich zuerst.",
+      "wrongUserOrPw": "Falsche Email Adresse oder falsches Passwort angegeben.",
+      "enterEmailForReset": "Bitte fülle das E-Mail Feld aus, um dein Passwort zurück zu setzen.",
+      "pwResetError": "Fehler beim Passwort zurücksetzen. Bist du sicher, dass es seine E-Mail ist?",
```
florianschmidt1994: seine → deine. Maybe we could use a whole different sentence here, I don't find this super clear. 
Suggestion "Fehler beim Zurücksetzen des Passworts. Bist du sicher, dass du deine E-Mail korrekt angegeben hast?"

src/App.jsx
```diff
@@ -101,6 +102,12 @@ function App(props) {
                 <Route path="/signup">
```
florianschmidt1994: Do we have / need support for :returnUrl here as well?

src/views/Signin.jsx
```diff
@@ -0,0 +1,131 @@
...
+      switch (signInResult.code) {
+        case 'auth/user-not-found': return setError(t('views.signIn.noUser'));
+        case 'auth/wrong-password': return setError(t('views.signIn.wrongUserOrPw'));
+        default: return setError(signInResult.message);
+      }
+    }
+    if (!signInResult.user.emailVerified) await signInResult.user.sendEmailVerification();
+  };
+
+  // eslint-disable-next-line consistent-return
```
florianschmidt1994: We should investigate what the proper way to do this is

src/views/Signup.jsx
```diff
@@ -33,62 +33,77 @@ const Signup = (props) => {
...
-  const sendPasswordResetMail = async (e) => {
-    e.preventDefault();
-    if (!email) return setError('Bitte fülle das E-Mail Feld aus, um dein Passwort zurück zu setzen.');
-    fb.auth.sendPasswordResetEmail(email, {
-      url: `${baseUrl}/#/signup`,
-      handleCodeInApp: false,
-    })
-      .then(() => setPasswordResetSuccess(true))
-      .catch(() => setError('Fehler beim Passwort zurücksetzen. Bist du sicher, dass es deine E-Mail ist?'));
+  const comparePasswords = () => {
```
florianschmidt1994: Do we need the ref here? Could we express this with useEffect or similar?

src/App.jsx
Maurice22: Wes we do have it here as well the route i defined directly above but was probably hidden as it was no change from before.

src/views/Signin.jsx
Maurice22: So, the issue is the following: As soon as you return a value inside a lambda function ones, Eslint expects you to return a value for all code paths. As we return the call to the `setError` function in line 61, eslint will mark that we don't return any value in case the `(!email)` condition is false.

2 ways to solve this:
1. Separate `setError()` and `return` so that we return void instead of the function call
2. Add an else statement to the condition and don't return anything

florianschmidt1994: I would vote for #1. Reason: I prefer early return over large if / else blocks, as I believe they are easier to read

```
  const sendPasswordResetMail = async (e) => {
    e.preventDefault();
    setError('');
    if (!email) {
      setError(t('views.signIn.enterEmailForReset'));
      return;
    }

    fb.auth.sendPasswordResetEmail(email, {
      url: `${baseUrl}/#/signin`,
      handleCodeInApp: false,
    })
      .then(() => setPasswordResetSuccess(true))
      .catch(() => setError(t('views.signIn.pwResetError')));
  };
```

src/views/Signup.jsx
Maurice22: No, as we need the reference to the input to set the custom validation error message

Maurice22:
> The alert which indicates that a password reset email has been sent is not very prominent and I nearly missed it in my review, maybe we can give a little more focus to this or maybe even put it where the current warnings for other cases also are?

Moved it to where the errors are shown to be closer to the actual button.

> With HTML5 form validation turned off, when entering an invalid email there is an error printed to the screen (as expected!) In addition, there is also an uncaught exception in the code

I added an additional custom error message for this case but I can't get rid of the console error (even with a try catch) and I think it is fine if we just shot the custom error and don't add an additional validation below the HTML5 form validation

> Chrome prompts me to add autocomplete to the repeat password field

Added the `autocomplete="off"` attribute to the password repeat field to encourage actual retyping of the password.

> "What about testes?"

On it ;)

florianschmidt1994:
Looks good to me! I've replaced my personal email with some placeholder emails. After we got a second pair of eyes on this I will merge this (and resolve the conflicts)

tgraupne:
I am on it...

public/locales/de/translation.json
```diff
@@ -154,6 +154,39 @@
...
+      "yourPw": "Dein Passwort",
+      "email": "Email",
+      "password": "Passwort",
+      "resetPw": "Passwort zurücksetzen",
+      "loginNow": "Jetzt anmelden",
+      "registerNow": "Neu registrieren",
+      "pwResetConfirmation": "Eine Email mit Anleitung zum Zurücksetzen deines Passworts wurde dir zugesendet!",
+      "headerText": "Melde dich mit deiner E-Mail und deinem Passwort an um {{reasonForSignin}}.",
+      "reasonForSigninDefault": "eine Hilfe-Anfrage zu posten",
+      "emailInvalid": "Bitte gib deine korrekte Emailadresse an."
```
tgraupne: During my tests I watched the `too-many-requests` error and I thought it could be a good idea to catch that as well.

```suggestion
      "emailInvalid": "Bitte gib deine korrekte Emailadresse an.",
      "tooManyRequests": "Das waren zu viele Fehlversuche, bitte versuche es später noch einmal."
```

src/views/Signin.jsx
```diff
@@ -0,0 +1,138 @@
...
+
+  const signIn = async (e) => {
+    e.preventDefault();
+    const signInResult = await signInWithEmailAndPassword(email, password);
+    if (signInResult.code) {
+      switch (signInResult.code) {
+        case 'auth/user-not-found': setError(t('views.signIn.noUser')); break;
+        case 'auth/wrong-password': setError(t('views.signIn.wrongUserOrPw')); break;
+        case 'auth/invalid-email': setError(t('views.signIn.emailInvalid')); break;
+        default: setError(signInResult.message);
```
tgraupne: During my tests I watched the too-many-requests error and I thought it could be a good idea to catch that as well. 

```suggestion
      switch (signInResult.code) {
        case 'auth/user-not-found': setError(t('views.signIn.noUser')); break;
        case 'auth/wrong-password': setError(t('views.signIn.wrongUserOrPw')); break;
        case 'auth/invalid-email': setError(t('views.signIn.emailInvalid')); break;
        case 'auth/too-many-requests': setError(t('views.signIn.tooManyRequests')); break;
        default: setError(signInResult.message);
```

tgraupne:
But srsly @Maurice22, great work with the test coverage of this new feature. 👍

Issue #158, opened 2020-03-26
Discuss usage of error reporting tool


florianschmidt1994:
With more traffic and users, we have an increase of reports where stuff is not working properly. We should discuss the usage of an error monitoring tool such as Sentry. Maybe Firebase has something like this integrated as well?

- [ ] Investigate error reporting tools
- [ ] Discuss GDPR implications
- [ ] Discuss if the cost / benefit is worth adding something like this in the first place

Feedback appreciated! @kenodressel @tgraupne

Maurice22:
I checked out Sentry and it was really able to set up. It only gets activate if the user accepted cookies in my current version. I'm also in contact with them about getting a free plan as we are open source.

florianschmidt1994:
Nice! Should we go with the free version for now / trail for 14 days as we wait for feedback? What do you think?

Maurice22:
Sure, it's really only 4 lines of code or so in my local branch wo I will activate the trial and create a PR for this

PR #177, opened 2020-03-28
Feature/error reporting

Maurice22:

Adds error reporting as discussed in #158 .

Adds sentry and links it to the QH project. Added manual error reporting at places we currently catch errors already.

**Checklist**
- [ ] `yarn build` passes
- [ ] `yarn lint` does not show any errors

**Others details**
- [x] This PR introduces a new dependency: @sentry/browser required for the sentry error reporting


src/App.jsx
```diff
@@ -42,7 +43,11 @@ function App(props) {
   } = props;
 
 
-  const addListener = () => {
+  const enableAnalytics = () => {
+    // Crash reportin
+    Sentry.init({ dsn: 'https://d00e781dba684e6e93070bcb8f2da4c6@sentry.io/5178489' });
```
florianschmidt1994: Can we use .env files for this?

src/views/NotifyMe.jsx
```diff
@@ -29,6 +30,7 @@ export default function NotifyMe() {
 
       setSignInLinkSent(true);
     } catch (error) {
+      Sentry.captureException(error);
```
florianschmidt1994: What happens here if the cookie consent has not been accepted?

Maurice22: No errors are shown and the errors are also not written into a backlog. So it's basically a noop.

src/App.jsx
```diff
@@ -42,7 +43,11 @@ function App(props) {
   } = props;
 
 
-  const addListener = () => {
+  const enableAnalytics = () => {
+    // Crash reportin
```
jagaSto: typo, should be 
```suggestion
    // Crash reporting
```

:)

Maurice22:
Good point, switched to environments instead of projecs

florianschmidt1994:
Looks good, will merge!

PR #183, opened 2020-03-28
Scroll to the right position when menu bar visible

Maurice22:

Scrolling down to thee right position on small screens now by adding a conditional offset based on the media query. The underlying issue is that we show the menu bar on mobile but not on desktop, so on mobile it overlaps the buttons.

**Checklist**
- [ ] `yarn build` passes
- [ ] `yarn lint` does not show any errors

### DIFF
--- a/src/views/Main.jsx
+++ b/src/views/Main.jsx
@@ -12,7 +12,7 @@ export default function Main() {
   const { t } = useTranslation();
 
   function scrollTo() {
-    const yOffset = -30;
+    const yOffset = window.matchMedia('(min-width: 835px)').matches ? -30 : -80;
     const el = document.getElementById('hilfe-buttons');
     if (el) {
       const y = el.getBoundingClientRect().top + window.pageYOffset + yOffset;
--- a/src/views/Main.jsx
+++ b/src/views/Main.jsx
@@ -12,7 +12,7 @@ export default function Main() {
   const { t } = useTranslation();
 
   function scrollTo() {
-    const yOffset = window.matchMedia('(min-width: 835px)').matches ? -30 : -80;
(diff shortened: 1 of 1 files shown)


src/views/Main.jsx
```diff
@@ -12,7 +12,7 @@ export default function Main() {
   const { t } = useTranslation();
 
   function scrollTo() {
-    const yOffset = -30;
+    const yOffset = window.matchMedia('(min-width: 835px)').matches ? -30 : -80;
```
Maurice22: Any chance we can make this dynamic by reading the tailwind config?

tgraupne:
Unfortunately, I don't quite get the misbehaviour. @Maurice22, could you get in touch with me and send me screenshot?

kenodressel:
@Maurice22 were you able to find the implementation by @timjb in the git history?

Maurice22:
The only prior version to my initial changes I found is
```
function scrollTo() {
    const el = document.getElementById("anfragen");
    if (el) {
      el.scrollIntoView({behavior: 'smooth'})
    }
  }
```

Which was done by @kenodressel

Maurice22:
> Unfortunately, I don't quite get the misbehaviour. @Maurice22, could you get in touch with me and send me screenshot?

If you open the site on a phone and click on the "Scroll Down" arrow, the call to action buttons will be partly coverd by the top menu bar.

kenodressel:
In terms of dynamic reading from tailwind config its possible: https://stackoverflow.com/questions/59982018/how-do-i-get-tailwinds-active-breakpoint-in-javascript but not sure its worth it to bloat the bundle with this. Additionally I was wrong in terms of the previous implementation I remembered it differently.

Additionally I would use `window.innerWidth >= 835`. Just looks better IMO.

Issue #182, opened 2020-03-28
FAQ Update


tgraupne:
We reviewed our FAQ page internally and created an updated. Those updates should end up online.

tgraupne:
I'll wait for PR #152 to start working on this because this update highly involves working on the translation files.

PR #193, opened 2020-03-30
Updated FAQ component with a new logic and new content

tgraupne:

This PR introduces solves #182 by updating the FAQ component.
Logic and have been updated.

**Checklist**
- [x] `yarn build` passes
  
For review ping anyone in this list:
 
 - @florianschmidt1994
 - @kenodressel
 - @Maurice22
 -   @tgraupne


public/locales/de/translation.json
```diff
@@ -91,42 +91,74 @@
...
+        "howDoesItWork": "Auf der Plattform quarantänehelden.org kommen Menschen zusammen, die sich gegenseitig unterstützen. Hier findest Du Anfragen von Personen, die etwas benötigen, aber selbst ihr Zuhause oder ihren Quarantäne-Ort nicht verlassen können. Als Hilfesuchende*r kannst Du hier Anfragen einstellen und Menschen in Deiner Nähe werden informiert.\n\nAls Helfende*r  kannst Du dir Anfragen in Deiner Nähe anzeigen lassen und diese beantworten. Wenn Du Dich mit Deiner E-Mail Adresse anmeldest und angibst, in welchem Umkreis Du Deine Hilfe anbieten möchtest, wirst Du von uns automatisch per Mail benachrichtigt, sobald jemand Hilfe benötigt. Wir leiten Deine Nachricht an den/die Anfragensteller*in weiter, die dann zu Dir Kontakt aufnehmen kann. Alle weiteren Details zur Anfrage (was wird benötigt, wann und wo?) werden dann direkt und persönlich geklärt.",
+        "whereDoesItWork": "Quarantänehelden.org ist im gesamten DACH-Raum verfügbar. Du kannst also Hilfegesuche in Deutschland, Österreich und der Schweiz aufgeben und beantworten.",
```
kenodressel: Should we also include "Süd-Tirol"?

public/locales/de/translation.json
```diff
@@ -91,42 +91,74 @@
...
-        "howContact": "Helfende können www.quarantaenehelden.org nach Anfragen in ihrer Nähe suchen. Die ausgewählte Anfrage wird angeklickt. Über ein Formular geht eine Email an die Person in Quarantäne. Sie erhält so eine Nachricht von dir und ihr könnt alles weitere verabreden.",
-        "whatToConsider": "Bitte achtet bei einer Übergabe darauf, dass es keinen Kontakt zwischen einer Person in Quarantäne und der helfenden Person gibt. Ihr könnt zum Beispiel etwas vor die Tür legen und dann anrufen, dass es da ist.",
+        "canICreateRequestsForOthers": "Ja, Du kannst auch Anfragen stellvertretend für jemand anderen stellen. Viele Nutzer*innen von quarantänehelden.org stellen beispielsweise Anfragen für ihre Großeltern, die weiter weg wohnen. Die initiale Kontaktaufnahme läuft in diesem Fall über die angegebene E-Mail-Adresse, danach können Helfer*Innen direkt mit den Hilfesuchenden Kontakt aufnehmen.",
+        "canICreateRequestsForMultipleAreas": "Ja, da Du auch Anfragen stellvertretend für jemand anderen stellen kannst, kannst Du auch für verschiedene Orte Hilfsanfragen einstellen. Die initiale Kontaktaufnahme läuft dann ebenfalls über die hinterlegte E-Mail-Adresse.",
```
kenodressel: Should we mention that you automatically notify people in surround areas? Because this could encourage people to create the same request multiple times with slightly different postal codes.

public/locales/de/translation.json
```diff
@@ -91,42 +91,74 @@
...
+        "howToPay": "Handelt es sich bei der Anfrage um einen Einkauf, könnt Ihr selbst absprechen, wie Ihr die Bezahlung des Einkaufs handhabt. Gleichwohl solltet Ihr im Vorhinein absprechen, wie die Übergabe kontaktlos ablaufen kann",
```
kenodressel: missing `.` at the end

public/locales/de/translation.json
```diff
@@ -91,42 +91,74 @@
...
+          "url": "/"
+        },
+        "howHelpPeopleWithoutInternet": {
+          "preLink": "Du kannst unseren Aushang im Treppenhaus deiner Wohnung aufhängen. Vorlagen in mehreren Sprachen haben wir dir ",
           "link": "auf der Startseite",
-          "postLink": "bereitgestellt."
+          "postLink": " bereitgestellt.",
+          "url": "/"
+        },
+        "whoAreYou": "Wir sind Andy, Florian, Henrike, Jakob, Julia, Keno, Nicolai, Milena, Philipp, Maurice, Janik, Alisa, Christina, Tom, Denise, Domenic und Sophie, eine Gruppe von Freunden, die mit diesem Projekt ihren kleinen Beitrag leisten und Menschen helfen möchten.",
```
kenodressel: I initially did an alphabetic sorting. I would love to continue this :)

src/views/FAQ.jsx
```diff
@@ -1,63 +1,74 @@
...
-        {t('views.faq.howHelpPeopleWithoutInternet.answer.preLink')}
-        <Link to="/" className="text-secondary hover:underline">
-          {' '}
-          {t('views.faq.howHelpPeopleWithoutInternet.answer.link')}
-          {' '}
-        </Link>
-        {t('views.faq.howHelpPeopleWithoutInternet.answer.postLink')}
-      </QA>
+  function buildFAQ(translationString) {
+    if (i18n.exists(`views.faq.answers.${translationString}.preLink`)) {
```
kenodressel: we could use `{ condition ? <if true> : <if false> }` this would also remove the requirement for `/* eslint-disable no-else-return */`

public/locales/de/translation.json
tgraupne: I think this depends on how we move forward with using the google auto-complete-api. If we for example decide that this is too expensive and we should build a cache or our own system based on lists of zipcode <-> lat/lng, than maybe adding a new country is too much.

Apart from that, with the current system, we could totally add "Süd-Tirol" as well.

public/locales/de/translation.json
tgraupne: Good point, do wo have any data on how much people did that already?

public/locales/de/translation.json
tgraupne:
```suggestion
        "howToPay": "Handelt es sich bei der Anfrage um einen Einkauf, könnt Ihr selbst absprechen, wie Ihr die Bezahlung des Einkaufs handhabt. Gleichwohl solltet Ihr im Vorhinein absprechen, wie die Übergabe kontaktlos ablaufen kann.",
```

public/locales/de/translation.json
tgraupne: Alright, I'll reshuffle things once we resolved all the other comments, ok?

public/locales/de/translation.json
kenodressel: To be honest I think südtirol is already enabled in the current version of autocomplete, check https://github.com/kenodressel/quarantine-hero/blob/master/src/components/LocationInput.jsx#L88 for reference. Additionally we would need to add a country field anyways if we disable the google autocomplete

public/locales/de/translation.json
kenodressel: I have deleted various posts in the past, from my gut I would say we had around 5-10 cases.

public/locales/de/translation.json
tgraupne: Alright, I added Süd-Tirol to the list.

public/locales/de/translation.json
tgraupne: I think it is not necessary to add that because people who ask for help don't really need to know how our "algorithm" works from the FAQs, right?

src/views/FAQ.jsx
```diff
@@ -1,63 +1,45 @@
+/* eslint-disable no-else-return */
```
kenodressel: this should not be needed anymore otherwise looks good

tgraupne: I am on it

PR #195, opened 2020-03-31
updates deploy scripts

kenodressel:

- adds firebase to deploy
- separates build and deploy steps
- adds linting to building
- building now also runs on PRs to master

**Checklist**
- [x] `yarn build` passes


.github/workflows/build-firebase.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+cd ./firebase/functions && \
```
florianschmidt1994: Will the scripts be run from the project root?

kenodressel: yes the scripts are always running from project root

kenodressel:
I think it should work but the real test comes once we move it to the master branch and the build scripts are run, this might require an additional PR to get those errors fixed

PR #196, opened 2020-03-31
Refactor firebase functions

kenodressel:

- moves scheduled function to eu servers
- runs sendEmails every 3 minutes between 9:00 and 24:00 Berlin Time

**Additonal Info**
- git diff seems weird, only changed line 98 to 98 - 100


florianschmidt1994:
I've renamed this PR because I've additionally refactored the firebase functions by moving the callback handlers into separate functions. Playing the ball back to you @kenodressel, your original changes looked good to me, how about mine? ;)

kenodressel:
I agree, changes look fine, can't approve since its my own pr but clear for merge.

PR #198, opened 2020-03-31
add redirect to verify-email if email is not verified

### LINKED ISSUE #197: Unverified users can attempt to create an ask-for-help-entry
* showcase: https://imgur.com/a/8xbtQJ3

## Steps to reproduce

* create a new account
* do not verify the email address
* try to create a new **ask-for-help** listing

## Severity

* affects every new user who tries to create a listing after immediate signup

jagaSto:

* fixes #197 (this is currently only happening on staging, since the firebase functions on prod have been updated yet)
* currently, a user with an unverified email can attempt to create a new listing, even though he should verify his email first
* the listing creation itself fails due to the security rules, which is fine
* additionally, the listing creation fails in the first session for a user even though if the user is verified
* the reason for this is the fact that the firebase backend is not aware yet that the user email is verified, for this we need to force refresh the ID token https://stackoverflow.com/questions/47243702/firebase-token-email-verified-going-weird

_Briefly describe your changes here_
* adds a redirect on the `ask-for-help` view to the `verify-email` view, if the user has not verified their email yet
* force refreshes the auth ID token on email verification

### DIFF
--- a/src/views/AskForHelp.jsx
+++ b/src/views/AskForHelp.jsx
@@ -69,11 +69,11 @@ export default function AskForHelp() {
     setPlaceId(pId);
   };
 
-  if (!isAuthLoading && (!user || !user.email)) {
+  if (!user || (!isAuthLoading && (!user || !user.email))) {
     return <Redirect to="/signin/ask-for-help" />;
   }
 
-  if (!user || !user.emailVerified) {
+  if (!user.emailVerified) {
     return <Redirect to="/verify-email" />;
   }
 
--- a/src/views/AskForHelp.jsx
+++ b/src/views/AskForHelp.jsx
@@ -73,6 +73,10 @@ export default function AskForHelp() {
     return <Redirect to="/signin/ask-for-help" />;
   }
 
+  if (!user || !user.emailVerified) {
+    return <Redirect to="/verify-email" />;
+  }
+
   return (
     <form onSubmit={handleSubmit} className="p-4">
       <h1 className="font-teaser py-4 pt-10">{t('views.askForHelp.createRequest')}</h1>
--- a/src/views/VerifyEmail.jsx
+++ b/src/views/VerifyEmail.jsx
@@ -21,8 +21,15 @@ const VerifyEmail = (props) => {
     setSendVerificationSuccess(true);
   };
 
-  const recheckEmail = (e) => {
+  const recheckEmail = async (e) => {
     e.preventDefault();
+    // https://github.com/firebase/firebase-js-sdk/issues/2529
+    // https://stackoverflow.com/questions/47243702/firebase-token-email-verified-going-weird
+    // after an immediate sign up and email verification, "firebase.auth().currentUser.emailVerified" is set correctly
+    // however, "auth.token.email_verified" in the firebase backend gets ist value from the ID token which will not get updated until it expires or it is force refreshed
+    // therefore, we need to force refresh the ID token, to make the backend aware that the user is indeed verified, and passes the security rules
+    // by design, if the user would log out and log in again, the token would be refreshed as well
+    await firebaseAppAuth.currentUser.getIdToken(true);
     window.location.reload();
   };
 


Maurice22:
Do we maybe want to make this a global guard to all sorts of places that requires a valid login? (i.e dashboard, ask-for-help, offer help and somewhat also when a user wants to report a post)?

src/views/AskForHelp.jsx
```diff
@@ -73,6 +73,10 @@ export default function AskForHelp() {
     return <Redirect to="/signin/ask-for-help" />;
   }
 
+  if (!user || !user.emailVerified) {
```
kenodressel: if not user should we redirect to login?

src/views/VerifyEmail.jsx
```diff
@@ -21,8 +21,15 @@ const VerifyEmail = (props) => {
...
+  const recheckEmail = async (e) => {
     e.preventDefault();
+    // https://github.com/firebase/firebase-js-sdk/issues/2529
+    // https://stackoverflow.com/questions/47243702/firebase-token-email-verified-going-weird
+    // after an immediate sign up and email verification, "firebase.auth().currentUser.emailVerified" is set correctly
+    // however, "auth.token.email_verified" in the firebase backend gets ist value from the ID token which will not get updated until it expires or it is force refreshed
+    // therefore, we need to force refresh the ID token, to make the backend aware that the user is indeed verified, and passes the security rules
+    // by design, if the user would log out and log in again, the token would be refreshed as well
+    await firebaseAppAuth.currentUser.getIdToken(true);
     window.location.reload();
```
kenodressel: isn't the getIdToken exactly to avoid reloading the page?

src/views/AskForHelp.jsx
jagaSto: line https://github.com/kenodressel/quarantine-hero/blob/adc108871814562713f15ddfe18c0b1d40c3a950/src/views/AskForHelp.jsx#L72 is taking care of that, will remove the `!user` check

src/views/VerifyEmail.jsx
jagaSto: yes and no, reloading alone is not enough to force refresh the token, but we could probably do without the `window.location.reload();`. Shall test and get back to you

src/views/AskForHelp.jsx
jagaSto: nvm user can still be undefined at that point (if user is not is logged in) --> check needs to stay

src/views/VerifyEmail.jsx
jagaSto: nope, we need to reload to fetch the user auth state again.

kenodressel: Not sure this is the optimal way but since this is a hotfix I think we go for it anyways

kenodressel:
@Maurice22 I agree but I say we open an issue and merge this hotfix for now.

src/views/AskForHelp.jsx
florianschmidt1994, line 76: I've tested this locally and if I'm logged out I'll be redirected to `/#/verify-email`. This is not expected behaviour, right?

jagaSto: good catch, definitely not, will fix!

jagaSto: @florianschmidt1994 you may try again :)

florianschmidt1994:
Merging

src/views/AskForHelp.jsx
```diff
@@ -69,10 +69,14 @@ export default function AskForHelp() {
     setPlaceId(pId);
   };
 
-  if (!isAuthLoading && (!user || !user.email)) {
+  if (!user || (!isAuthLoading && (!user || !user.email))) {
```
Maurice22, line 72: The latter !user will always be false so I guess the expression can be refined
```suggestion
  if (!user || (!isAuthLoading && !user.email)) {
```

Issue #173, opened 2020-03-27
Error in Auth dependency


kenodressel:
We currently have `react-with-firebase-auth` and `react-firebase-hooks` and use mostly the latter package. Additionally we have possible errors linked to `react-with-firebase-auth`:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in withFirebaseAuth(Signin) (at App.jsx:106)
console.<computed> @ 0.chunk.js:149557
```

florianschmidt1994:
I want to wait for https://github.com/kenodressel/quarantine-hero/pull/168 to be merged, because it touches the same parts

PR #203, opened 2020-04-02
Fix/173 remove errors in auth dependencies

florianschmidt1994:
This change replaces the withFirebaseAuth higher order component with the `useAuthState()` hook that we use in most of the places anyway.

This PR includes
- Replacing HoC with hook in App
- Replacing HoC with hook in Signin
- Replacing HoC with hook in SignUp
- Replacing HoC with hook in VerifyEmail
- Adds test coverage for navigation in landingpage
- Adds test coverage for verifyEmail
- Changes behaviour in verifyEmail. Before: signed in and verified user would see the page. After: signed in and verified user will be redirected to /#/ask-for-help

I've originally approached these changes in order to remove the warning from #173, but I've figured once I'm at it I could even remove all the HoC occurrences as well.

**Others details**
- [x] This PR removes a dependency. Reason: react-with-firebase-auth is no longer needed

### DIFF
--- a/cypress/integration/verifyEmail.spec.js
+++ b/cypress/integration/verifyEmail.spec.js
@@ -1,0 +1,43 @@
+context('Verify Email Page', () => {
+
+  describe('User is not logged in', () => {
+    beforeEach(() => {
+      indexedDB.deleteDatabase('firebaseLocalStorageDb');
+      cy.visit('localhost:3000/#/verify-email');
+    });
+
+    it('visiting the page should redirect to sign up', () => {
+      cy.hash().should('equal', '#/signup');
+    });
+  });
+
+  describe('User is logged in and his email address has not been verified', () => {
+
+    beforeEach(() => {
+      indexedDB.deleteDatabase('firebaseLocalStorageDb');
+      cy.visit('localhost:3000/#/signin');
+      cy.get('form input[type="email"]').type('not.verified@example.com{enter}');
+      cy.get('form input[type="password"]').type('test1234{enter}');
+    });
+
+    it('visiting the page should not redirect him', () => {
+      cy.visit('localhost:3000/#/verify-email');
+      cy.hash().should('equal', '#/verify-email');
+    });
+  });
+
+  describe('User is logged in and his email address has been verified', () => {
+
+    beforeEach(() => {
+      indexedDB.deleteDatabase('firebaseLocalStorageDb');
+      cy.visit('localhost:3000/#/signin');
+      cy.get('form input[type="email"]').type('verified@example.com{enter}');
+      cy.get('form input[type="password"]').type('test1234{enter}');
+    });
+
+    it('visiting the page should redirect to ask-for-help', () => {
+      cy.visit('localhost:3000/#/verify-email');
+      cy.hash().should('equal', '#/ask-for-help');
+    });
+  });
+});
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -223,4 +223,3 @@ export default function App() {
     </div>
   );
 }
-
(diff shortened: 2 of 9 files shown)


cypress/integration/verifyEmail.spec.js
```diff
@@ -0,0 +1,43 @@
...
+  describe('User is logged in and his email address has been verified', () => {
+
+    beforeEach(() => {
+      indexedDB.deleteDatabase('firebaseLocalStorageDb');
+      cy.visit('localhost:3000/#/signin');
+      cy.get('form input[type="email"]').type('verified@example.com{enter}');
+      cy.get('form input[type="password"]').type('test1234{enter}');
+    });
+
+    it('visiting the page should redirect to ask-for-help', () => {
```
jagaSto, line 38: nice!

src/App.jsx
```diff
@@ -75,18 +124,7 @@ function App(props) {
...
-          {user && (
-            <>
-              <Link className="mr-6 font-open-sans text-gray-700" to="/dashboard">{t('components.desktopMenu.myOverview')}</Link>
-              <button type="button" data-cy="btn-sign-out" className="mr-4 font-open-sans text-gray-700" to="#" onClick={signOut}>{t('components.desktopMenu.signOut')}</button>
-            </>
-          )}
-          <Link className="mr-6 font-open-sans text-gray-700" to="/press">{t('App.press')}</Link>
-          <ShareButtons />
-        </div>
+        <TopNavigation user={user} isAuthLoading={isAuthLoading} signOut={signOut} />
```
jagaSto, line 127: much easier to read what's going on now 👏

src/views/Signin.jsx
```diff
@@ -42,18 +39,18 @@ const Signin = (props) => {
 
   const signIn = async (e) => {
     e.preventDefault();
-    const signInResult = await signInWithEmailAndPassword(email, password);
-    if (signInResult.code) {
-      switch (signInResult.code) {
+    try {
+      const signInResult = await signInWithEmailAndPassword(email, password);
+      if (!signInResult.user.emailVerified) await signInResult.user.sendEmailVerification();
```
jagaSto, line 44: could `signInResult.user` ever be undefined here?

florianschmidt1994: according to the TS signature yes, unfortunately

```ts
  type UserCredential = {
    additionalUserInfo?: firebase.auth.AdditionalUserInfo | null;
    credential: firebase.auth.AuthCredential | null;
    operationType?: string | null;
    user: firebase.User | null;
  };
```

florianschmidt1994: I've added some code to cover this case and report an error to sentry. The firebase documentation is not very clear why this might ever happen, but at least we can capture it this way

jagaSto: 👍

src/App.jsx
```diff
@@ -75,18 +123,7 @@ function App(props) {
...
```
Maurice22, line 126: Do we need to pass isAuthLoading here? For me it seems that user will be undefined till authentication is no longer loading, hence we could reduce the condition in the `Top Navigation` and remove the property

src/views/Signin.jsx
```diff
@@ -9,25 +8,24 @@ import {
...
 import Footer from '../components/Footer';
 import MailInput from '../components/MailInput';
 import fb from '../firebase';
-import { baseUrl } from '../appConfig';
 
-const firebaseAppAuth = firebaseApp.auth();
+import { baseUrl } from '../appConfig';
 
-const Signin = (props) => {
+export default () => {
```
Maurice22, line 19: Any particular reason why we are removing the function name here?

PR #204, opened 2020-04-03
Break words in entry if too long

### LINKED ISSUE #159: Linebreak broken on Safari, Mac
On Safari on mac it can happen that the text overflows the entry component because it is not broken at the right spot.

Steps to reproduce: Search for the entry shown in the screenshot

Bildschirmfoto 2020-03-26 um 11 34 16

florianschmidt1994:
Break words properly to avoid overflow

_Before_
image

_After_
image

I've verified this on mobile and desktop

Close #159

### DIFF
--- a/src/components/Entry.jsx
+++ b/src/components/Entry.jsx
@@ -161,7 +161,7 @@ export default function Entry(props) {
   const requestCard = (
     <Link
       to={entryBelongsToCurrentUser ? '/dashboard' : `/offer-help/${props.id}`}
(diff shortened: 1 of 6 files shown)


Maurice22:
Should we think about the css property hyphens with an appropriate hyphenation-resource. C.f. https://css-tricks.com/almanac/properties/h/hyphenate/ in order to add hyphens to long words that would otherwise just be broken at the end of the container?

florianschmidt1994:
_Before_
image

_After_
image

Thanks @Maurice22 I didn't know about this!

florianschmidt1994:
I've also moved the Entry.jsx in a folder called components/entry/ to add support for component css

PR #219, opened 2020-04-10
Added view for use in iframes

### LINKED ISSUE #155: Iframe of FilteredListView for the integration on other websites
Other websites asked us if we could provide an iframe of our list of asks-for-help.

Maurice22:

There is a new view that can be included on external websites as iframes. The following changes were made:

* Added a new router stage in index.jsx to allow rendering views outside the main QH website theme
* Added the FilteredListFrame view with wich includes
  * CTA buttons
  * Header
  * Entry Container for visualization and search
* The FilteredListFrame has code that disables the react link behavior and sets the target="_blank" attribute on all anchor elements so that whenever a user clicks on something it results in opening the QH website in a new tab.

Fixes #155.


src/index.jsx
```diff
@@ -17,7 +23,16 @@ import './i18n'; // Internalization
 ReactDOM.render(
   <Suspense fallback={<Loader />}>
     {/* Suspense is necessary for internalization */}
-    <App />
+    <Router>
```
florianschmidt1994, line 26: why here instead of App?

src/views/FilteredListFrame.jsx
```diff
@@ -0,0 +1,77 @@
...
+import { Trans, useTranslation } from 'react-i18next';
+import { Link } from 'react-router-dom';
+import { baseUrl } from '../appConfig';
+import EntryContainer from '../components/EntryContainer';
+
+export default function FilteredListView() {
+  const { t } = useTranslation();
+
+  useEffect(() => {
+    const makeExternalLinks = () => {
```
florianschmidt1994, line 11: This is quite hacky. I kinda like how simple it is, but in some sense it feels wrong. Maybe we can leave but, but let's discuss it

Maurice22: An alternative would be to add an additional to all prop to all views/components that will be passed to every <Link> and sets target to "_blank". As this would mean major refactoring, I went with this solution

src/index.jsx
Maurice22: Because in app the have the switch already below the top navigation. I'm not quite sure but maybe I could add 2 routers to the App but I thought this way it is cleaner. -> Everything outside the main design is handled in /index and everything that follows the main design (i.e. Top Nav visible, etc) is handled in /App

florianschmidt1994: Maybe we should refactor the navigation thing then? If we believe that this is out of scope, we could open an issue for it. I would prefer to have a single place with all routes in the end

Maurice22: I have moved the additional routing step to the App View now. Can you check this again?

src/views/FilteredListFrame.jsx
florianschmidt1994: All right, let's leave it this way for now

Maurice22:
The above mentioned behavior seems to be unrelated to this PR and issue #220 was created for it

src/views/FilteredListFrame.jsx
```diff
@@ -0,0 +1,77 @@
...
+    makeExternalLinks();
+
+    mutationObserver.observe(document, { attributes: false, childList: true, subtree: true });
+
+    return () => mutationObserver.disconnect();
+  }, []);
+
+  return (
+    <div className="flex items-center flex-col">
+      <div className="iframe-header w-full relative overflow-hidden">
```
tgraupne, line 35: Hi @Maurice22, is there the possibility to set in minimum width for the iframe?
My tests show, that about 630 px is the minimum width where the iframe looks good.

tailwind.config.js
```diff
@@ -4,10 +4,16 @@ module.exports = {
   separator: ':',
   theme: {
     screens: {
+      xs: '320px',
       sm: '640px',
       md: '835px',
       lg: '1024px',
       xl: '1280px',
+      xsi: { max: '319px' },
```
florianschmidt1994, line 12: As discussed in private chat, please check if you can use the existing screen sizes

PR #224, opened 2020-04-13
Add flag for enabling/disabling i18n debug logging

### LINKED ISSUE #181: Provide a config parameter for i18next debug logging
Just recently, we deactivated debug logging for i18next.
However, it might be a good idea to have those logs in testing scenarios.
Therefore, it would be great to introduce a parameter for this.

jakob-ed:

Adds a flag to the `.env*` files to turn debug logging from `i18next` on/off (closes #181). It's enabled by default now for the dev env and disabled for the prod env.


src/i18n.js
```diff
@@ -17,7 +17,7 @@ i18n
   .init({
     fallbackLng: 'de', // @TODO Change to English as soon as we have English translations.
     lng: 'de',
-    debug: false,
+    debug: process.env.REACT_APP_DEBUG_LOGGING === 'true',
```
jakob-ed, line 20: `dotenv` unfortunately parses everything to String, including Booleans...

.env
```diff
@@ -8,4 +8,5 @@ REACT_APP_FIREBASE_MESSAGING_SENDER_ID=634343616076
 REACT_APP_FIREBASE_APP_ID=1:634343616076:web:2aa2153615cd5123b38ee9
 REACT_APP_FIREBASE_MEASUREMENT_ID=G-2PBE89H418
 REACT_APP_SENTRY_DSN=https://d00e781dba684e6e93070bcb8f2da4c6@sentry.io/5178489
-REACT_APP_ENV=production
\ No newline at end of file
+REACT_APP_ENV=production
+REACT_APP_DEBUG_LOGGING=false
```
tgraupne, line 12:
```suggestion
REACT_APP_ENV=production
REACT_APP_DEBUG_LOGGING=false

```

.env.development
```diff
@@ -11,4 +11,5 @@ REACT_APP_FIREBASE_MESSAGING_SENDER_ID=855597135666
 REACT_APP_FIREBASE_APP_ID=1:855597135666:web:9bf52744911dd9d252c92b
 REACT_APP_FIREBASE_MEASUREMENT_ID=G-MM2TK770SV
 REACT_APP_SENTRY_DSN=https://d00e781dba684e6e93070bcb8f2da4c6@sentry.io/5178489
-REACT_APP_ENV=staging
\ No newline at end of file
+REACT_APP_ENV=staging
+REACT_APP_DEBUG_LOGGING=true
```
tgraupne, line 15:
```suggestion
REACT_APP_ENV=staging
REACT_APP_DEBUG_LOGGING=true

```

jakob-ed:
Haha I prefer newlines at the EOF as well but the files didn't have any before so I just stuck with it :)

.env.development
florianschmidt1994: I disagree with having this enabled in dev by default, b.c. for me I don't want to have the output enabled unless when I'm specifically debugging something related to i18n.

kenodressel:
Looks good. Lets discuss if we should have this enabled by default or not in the next dev call because I am in favour of it :)

src/i18n.js
Maurice22: We have the appConfig.jsx should we think about putting all config params in there and do the required type casting to prevent having this at several places in the code?

jakob-ed: Sounds like a good plan once more non-String config params are added

PR #225, opened 2020-04-13
Enhancement/221 refactor navigation

florianschmidt1994:
- Move page "frame" into a separate component
- Clean up some code
- Change routes to be self-contained
- Transform `<ScrollToTop />` component to `useScrollToTop` hook 
- Remove some nested divs that were not necessary to nest

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -32,7 +32,7 @@ import DesktopMenu from './components/DesktopMenu';
 import VerifyEmail from './views/VerifyEmail';
 import CompleteOfferHelp from './views/CompleteOfferHelp';
 import NotifyMe from './views/NotifyMe';
-import ScrollToTop from './components/ScrollToTop';
+import useScrollToTop from './components/ScrollToTop';
 import ShareButtons from './components/ShareButtons';
 import Press from './views/Press';
 import createEventListener from './util/createEventListener';
@@ -89,7 +89,10 @@ function TopNavigation({ isAuthLoading, user, signOut }) {
   );
 }
 
-export default function App() {
+const Frame = (props) => {
+  // always scroll page to top when changing the pathname
+  useScrollToTop();
+
   const { t } = useTranslation();
   const [user, isAuthLoading] = useAuthState(firebase.auth());
   const signOut = () => firebase.auth().signOut();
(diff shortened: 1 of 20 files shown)


src/App.jsx
```diff
@@ -120,105 +124,173 @@ export default function App() {
...
+        style={{ fontWeight: '600' }}
+      >
+        <img alt="logo" src={require('./assets/logo_invert.svg')} className="h-10" />
+      </Link>
+      <div>
+        <MenuIcon style={{ fontSize: '40px' }} className="text-gray-600" onClick={() => setMenuOpen(true)} />
+      </div>
+    </div>
+  );
+
```
Maurice22, line 144: Any reason for not defining this outside the page just like the DesktopTopNavigation?

src/App.jsx
```diff
@@ -120,105 +124,173 @@ export default function App() {
...
   return (
-    <div className="flex items-center min-h-screen flex-col bg-kaki">
-      <Router>
-        <TopNavigation user={user} isAuthLoading={isAuthLoading} signOut={signOut} />
+    <>
+      <div className="flex items-center min-h-screen flex-col bg-kaki">
+
+        <DesktopTopNavigation user={user} isAuthLoading={isAuthLoading} signOut={signOut} />
+        <MobileTopNavigation />
```
Maurice22, line 150: Should we use conditional rendering for this maybe?

src/App.jsx
```diff
@@ -120,105 +124,173 @@ export default function App() {
...
-                <Route component={NotFound} path="*" />
-              </Switch>
-              <ScrollUpButton
-                ContainerClassName="scroll-up-btn"
-                TransitionClassName="scroll-up-btn-fade"
-              >
-                <img alt="arrow-down" className="arrow-down" src={require('./assets/arrows_up.svg')} />
-              </ScrollUpButton>
-            </div>
+          <DesktopLowerNavigation isLoggedIn={user} signOut={signOut} />
```
Maurice22, line 153: As far as I can tell, we don't need tu pass those props anymore

src/App.jsx
florianschmidt1994: I thought about it, but that would mean that we need to check the viewport in javascript and I'd like to avoid it and leave it to css for nice responsiveness

florianschmidt1994:
@kenodressel I cannot reproduce this. Could you give me some more details? What browser & screensize?

Maurice22:
@florianschmidt1994 Should we think about putting the footer component into the new <Page> component instead of every view? I think the way it is now is error-prone if someone forgets the footer on a page

florianschmidt1994:
@Maurice22 done. Can you check it again?

PR #227, opened 2020-04-13
Made mobile the first test case for cypress and updated the github ac…

tgraupne:
By adding a typical mobile screen resolution to the cypress.json, we configured the testing to be mobile-first.

Moreover, we introduced new yarn commands to test mobile and desktop

### DIFF
--- a/.github/workflows/deploy-firebase.sh
+++ b/.github/workflows/deploy-firebase.sh
@@ -8,6 +8,5 @@ fi
 GIT_REV="$(git rev-parse HEAD)"&& \
 FIREBASE_PROJECT="quarantine-hero" && \
 FIREBASE_MESSAGE="QH $GIT_REV deployment to firebase" && \
-cd ./firebase/functions && \
 # deploy here because npm does not resolve environment variables in npm run
 npx firebase-tools deploy --only functions --project "$FIREBASE_PROJECT" -m "$FIREBASE_MESSAGE" --non-interactive
--- a/.github/workflows/deploy-firebase.sh
+++ b/.github/workflows/deploy-firebase.sh
@@ -1,12 +1,14 @@
 #!/usr/bin/env bash
+# exit script as soon as any command returns non-zero
+set -e
 
 if [[ -z "$FIREBASE_TOKEN" ]]; then
     echo "Set the FIREBASE_TOKEN env variable."
     exit 1
 fi
 
-GIT_REV="$(git rev-parse HEAD)"&& \
-FIREBASE_PROJECT="quarantine-hero" && \
-FIREBASE_MESSAGE="QH $GIT_REV deployment to firebase" && \
+GIT_REV="$(git rev-parse HEAD)"
+FIREBASE_PROJECT="quarantine-hero"
+FIREBASE_MESSAGE="QH $GIT_REV deployment to firebase"
 # deploy here because npm does not resolve environment variables in npm run
 npx firebase-tools deploy --only functions --project "$FIREBASE_PROJECT" -m "$FIREBASE_MESSAGE" --non-interactive
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -10,8 +10,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@master
+      - name: Install dependencies
+        run: yarn install
       - name: Build
-        run: ./.github/workflows/build-frontend.sh
+        run: yarn build
+        with:
+          env: NODE_ENV=prod
       - name: Deploy
         env:
           TOKEN: ${{ secrets.TOKEN }}
@@ -26,13 +30,16 @@ jobs:
           tagName: ${{ github.sha }}
           environment: production
 
-  deploy-firebase:
+  deploy-backend:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: firebase/functions
     steps:
       - name: Checkout
         uses: actions/checkout@master
-      - name: Build
-        run: ./.github/workflows/build-firebase.sh
+      - name: Install Dependencies
+        run: npm install
       - name: Deploy
         env:
           FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -30,7 +30,7 @@ jobs:
           tagName: ${{ github.sha }}
           environment: production
 
-  deploy-backend:
+  deploy-firebase:
     runs-on: ubuntu-latest
     defaults:
       run:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -54,7 +54,7 @@ jobs:
           wait-on: http://localhost:3000
           config-file: cypress-desktop.json
 
-  lint-backend:
+  lint-firebase:
     runs-on: ubuntu-latest
     defaults:
       run:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -36,9 +36,9 @@ jobs:
         uses: cypress-io/github-action@v1
         with:
           env: NODE_ENV=prod
-          build: yarn build
           start: yarn start
           wait-on: http://localhost:3000
+          config-file: cypress.json
   test-frontend-dekstop:
     runs-on: ubuntu-latest
     steps:
@@ -50,7 +50,6 @@ jobs:
         uses: cypress-io/github-action@v1
         with:
           env: NODE_ENV=prod
-          build: yarn build
           start: yarn start
           wait-on: http://localhost:3000
           config-file: cypress-desktop.json
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -5,18 +5,65 @@ on:
       - master
 
 jobs:
+  lint-frontend:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Install dependencies
+        run: yarn install
+      - name: Lint
+        run: yarn lint-frontend
   build-frontend:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@master
+      - name: Install Dependencies
+        run: yarn install
       - name: Build
-        run: ./.github/workflows/build-frontend.sh
-
-  build-firebase:
+        run: yarn build
+        env:
+          NODE_ENV: prod
+  test-frontend-mobile:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@master
-      - name: Build
-        run: ./.github/workflows/build-firebase.sh
+      - name: Install dependencies
+        run: yarn install
+      - name: Test frontend mobile
+        uses: cypress-io/github-action@v1
+        with:
+          env: NODE_ENV=prod
+          build: yarn build
+          start: yarn start
+          wait-on: http://localhost:3000
+  test-frontend-dekstop:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Install dependencies
+        run: yarn install
+      - name: Test frontend desktop
+        uses: cypress-io/github-action@v1
+        with:
+          env: NODE_ENV=prod
+          build: yarn build
+          start: yarn start
+          wait-on: http://localhost:3000
+          config-file: cypress-desktop.json
+
+  lint-backend:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: firebase/functions
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Install Dependencies
+        run: npm install
+      - name: Lint
+        run: npm run lint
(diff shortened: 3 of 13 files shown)


florianschmidt1994:
I would like to discuss whether we want to split the tests in two commands. Personally, I'm for leaving them in one command b.c. this gives contributers the confidence to say `yarn test` and if it passes, everything works (mobile & desktop). By splitting it up we are introducing a mechanism that is "not following the convention" and making it harder for others to contribute

tgraupne:
> I would like to discuss whether we want to split the tests in two commands. Personally, I'm for leaving them in one command b.c. this gives contributers the confidence to say `yarn test` and if it passes, everything works (mobile & desktop). By splitting it up we are introducing a mechanism that is "not following the convention" and making it harder for others to contribute

Well, couldn't we do something like:

```
{
  "name": "quarantine-hero",
  "version": "0.1.0",
  "private": true,
  ...
  "scripts": {
    "build:style": "tailwind build src/styles/index.css -o src/styles/tailwind.css",
    "start": "npm run build:style && react-scripts start",
    "build": "npm run build:style && react-scripts build",
    "test-mobile": "cypress run",
    "test-desktop":"cypress run --config viewportWidth=1280,viewportHeight=800",
    "test": "yarn test-mobile && yarn test-desktop",
    "eject": "react-scripts eject",
    "lint": "cd src && eslint . --ext .jsx .js --no-error-on-unmatched-pattern && cd ..",
    "lint-fix": "cd src && eslint . --fix --ext .jsx .js --no-error-on-unmatched-pattern && cd ..",
    "cypress": "cypress open"
  },
...
}
```

florianschmidt1994:
Yeah we could! I was wondering whether there is support for this built-in with cypress to give us output à la "user can login passed on desktop" "user can login failed on mobile", but that of course would just be a gimmick

.github/workflows/pr.yml
```diff
@@ -5,13 +5,44 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v1
+        with:
+          build: npm run build
+          start: npm start
```
Maurice22, line 44: @florianschmidt1994 do we need to run `npm run build` here as `npm start` will start and build anyways? Or should we maybe use another start command?

.github/workflows/pr.yml
```diff
@@ -5,13 +5,44 @@ on:
...
+        with:
+          node-version: 12.x
+      - name: npm install, build, and test
+        run: |
+          npm ci
+          npm run build --if-present
+          npm test
+        env:
+          CI: true
+          TERM: xterm
```
Maurice22, line 32: I think this is not necessary anymore as the job below works for running tests. I think we should rather set up a new job for linting. What do you think?

kenodressel: I would like to keep the master deploy as close as possible to the pr ci so I am not sure why we remove it from the build script at all.

.github/workflows/pr.yml
kenodressel: I am quite sure we can simply use `yarn start` here only. Should be verified through a commit tho.

.github/workflows/pr.yml
tgraupne: > I would like to keep the master deploy as close as possible to the pr ci so I am not sure why we remove it from the build script at all.

I agree, that the PR workflow should give us 100% confidence, that a deployment should work. However, that does not necessarily mean, that the deployment- and the pr-workflows have to be the same.

If we deploy on push to master and if we test and lint the the pr, we should not execute those to steps on a master push to safe some GitHub action execution minutes.

Moreover, after getting the tests in order, I would like to take a look at the deployment as well, because executing one bash script as a job is a lean approach, but in my opinion, it is not best practice.

My problem was, whatever I tried, I was stuck with the CONNECTIONERROR where cypress can't connect to 127.0.0.1:3000. If you could help me fix this, that would be terrific.

tgraupne:
image

All tests are working, we just ran into a Google API Limit for resetting passwords.
We should refactor this in a different issue 😅

.github/workflows/deploy-firebase.sh
```diff
@@ -8,6 +8,5 @@ fi
 GIT_REV="$(git rev-parse HEAD)"&& \
 FIREBASE_PROJECT="quarantine-hero" && \
 FIREBASE_MESSAGE="QH $GIT_REV deployment to firebase" && \
-cd ./firebase/functions && \
 # deploy here because npm does not resolve environment variables in npm run
 npx firebase-tools deploy --only functions --project "$FIREBASE_PROJECT" -m "$FIREBASE_MESSAGE" --non-interactive
```
Maurice22, line 12: Shouldn't we use yarn here as well?

.github/workflows/pr.yml
```diff
@@ -5,18 +5,65 @@ on:
       - master
 
 jobs:
+  lint-frontend:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Install dependencies
+        run: yarn install
```
Maurice22, line 14: Do we need to install the dependencies to lint?

.github/workflows/pr.yml
```diff
@@ -5,18 +5,65 @@ on:
...
+      - name: Test frontend mobile
+        uses: cypress-io/github-action@v1
+        with:
+          env: NODE_ENV=prod
+          build: yarn build
+          start: yarn start
+          wait-on: http://localhost:3000
```
Maurice22, line 41: Should we make the config file explicit here as well?

.github/workflows/pr.yml
```diff
@@ -5,18 +5,65 @@ on:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Install dependencies
+        run: yarn install
+      - name: Test frontend desktop
+        uses: cypress-io/github-action@v1
+        with:
+          env: NODE_ENV=prod
+          build: yarn build
```
Maurice22, line 53: I don't think wen need to add the build step here

.github/workflows/pr.yml
```diff
@@ -5,18 +5,65 @@ on:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Install dependencies
+        run: yarn install
+      - name: Test frontend mobile
+        uses: cypress-io/github-action@v1
+        with:
+          env: NODE_ENV=prod
+          build: yarn build
```
Maurice22, line 39: I don't think we need to add the build step here

.github/workflows/deploy-firebase.sh
tgraupne: Actually, I don't know. Since I am neither the expert on npm, nor on yarn, I'd like to ask @kenodressel or @florianschmidt1994 :-)

.github/workflows/pr.yml
tgraupne: I just tested with a "fresh" docker image of ubuntu:latest and since eslint is one of our dependencies, we need to install at least eslint.

At this point, especially with the prospect of caching, I wouldn't change `yarn install` to `something something install eslint ... `

.github/workflows/pr.yml
tgraupne: Same as above

.github/workflows/pr.yml
tgraupne: Because the build will get executed anyways by `yarn start`?

.github/workflows/pr.yml
tgraupne: Same as above

.github/workflows/pr.yml
Maurice22: Yes

.github/workflows/pr.yml
Maurice22: Watt do you mean?

Maurice22, line 41:
```suggestion
          wait-on: http://localhost:3000
          config-file: cypress.json
```

.github/workflows/pr.yml
```diff
@@ -5,18 +5,65 @@ on:
...
+        working-directory: firebase/functions
     steps:
       - name: Checkout
         uses: actions/checkout@master
-      - name: Build
-        run: ./.github/workflows/build-firebase.sh
+      - name: Install Dependencies
+        run: npm install
+      - name: Lint
+        run: npm run lint
```
Maurice22, line 69:
```suggestion
        run: npm run lint-backend
```

.github/workflows/deploy-firebase.sh
Maurice22: I guess in the end it does not really make a difference so we can leave it as is.

.github/workflows/pr.yml
tgraupne: Since we are providing a working directory and `lint-backend` is not part of the package.json within `firebase/functions`, `lint-backend` does not seem to work.

Hence, simply running `npm run lint` seems ok here.
What do you think @Maurice22

.github/workflows/pr.yml
tgraupne: I agree and I'll nice it a try.

.github/workflows/deploy-firebase.sh
```diff
@@ -8,6 +8,5 @@ fi
 GIT_REV="$(git rev-parse HEAD)"&& \
```
florianschmidt1994, line 8: My bash knowledge is a little rusty, why are we chaining the commands with `&& \` here instead of just separating them by newline?

.github/workflows/pr.yml
```diff
@@ -5,18 +5,64 @@ on:
...
+      - name: Install dependencies
+        run: yarn install
+      - name: Test frontend mobile
+        uses: cypress-io/github-action@v1
+        with:
+          env: NODE_ENV=prod
+          start: yarn start
+          wait-on: http://localhost:3000
+          config-file: cypress.json
+  test-frontend-dekstop:
```
florianschmidt1994, line 42: Should be `desktop`

.github/workflows/deploy-firebase.sh
Maurice22: AFAIK this way the execution will stop as soon as one command exists with a non-zero exit code

.github/workflows/pr.yml
Maurice22, line 42:
```suggestion
  test-frontend-desktop:
```

firebase/functions/package.json
```diff
@@ -7,7 +7,9 @@
     "start": "npm run shell",
     "deploy": "firebase deploy --only functions",
     "logs": "firebase functions:log",
-    "lint": "eslint ."
+    "test": "echo 'TODO'",
+    "lint": "eslint . --ext .jsx .js --no-error-on-unmatched-pattern",
```
florianschmidt1994, line 11: This did not work for me unfortunately. I've introduced a linting error in my code but it didn't show up in the output. Could you double-check?

cypress.json
```diff
@@ -1 +1,7 @@
-{}
+{
+  "viewportWidth": 375,
+  "viewportHeight": 812,
```
florianschmidt1994, line 3: Is this equal to some specific device and have you cross-checked with our tailwindcss breakpoints?

.github/workflows/deploy-firebase.sh
florianschmidt1994: We could also use 
```
# exit when any command fails
set -e
```

in that case, right?

firebase/functions/package.json
Maurice22: So it seems as if we add --no-error-on-unmatched-pattern the command will rust fail silently. I removed all the annotations now which makes eslint simply run on all .js files, which is fine for the backend

.github/workflows/deploy-firebase.sh
Maurice22: I assume so

tgraupne: Hi @florianschmidt1994, that is an excellent idea.
I am on it.

cypress.json
tgraupne: Hi @florianschmidt1994, this resolution is equal to an iPhone X.
Unfortunately, I did not think of cross-checking it with tailwind.
What would you suggest here, could you help me with that?

Maurice22: The size results in the breakpoint XS. Which seems fine for me as our main breakpoint between mobile and desktop is MD.

tgraupne: Sounds legit to me 😅

.github/workflows/deploy.yaml
```diff
@@ -26,13 +30,16 @@ jobs:
           tagName: ${{ github.sha }}
           environment: production
 
-  deploy-firebase:
+  deploy-backend:
```
kenodressel, line 33: why call it backend? I think firebase functions is a better naming

cypress.json
```diff
@@ -1 +1,7 @@
...
+  "env":{
+    "VIEWPORT": "mobile"
+  }
```
kenodressel, line 6: I think we should also disable screenshots & videos on failures. 
https://docs.cypress.io/guides/guides/screenshots-and-videos.html#Screenshots

Issue #213, opened 2020-04-06
Find a good way to track various z indices


florianschmidt1994:
We've recently merged #211, we should consider finding a good way to track the various z-index across our app. (@Maurice22 ;) )

tgraupne:
Won't do (for now)

florianschmidt1994:
I've re-opened this as the issue arose again in a PR review (#230)

mauriceackel:
Closing due to inactivity

PR #230, opened 2020-04-15
Refactor cookie banner

florianschmidt1994:
- Move Routes into separate component
- Move Page into separate component
- Remove react-cookie-consent dependency
- Implement cookie consent with React Context. They information about whether or not cookies are consented can now be accessed from any component using the React Context
- Implement sentry loading as component that will check cookie consent
- Use a hook for logging page changes if cookies are consented


Maurice22:
@florianschmidt1994 #225 just got merged, FYI

src/util/Sentry.jsx
```diff
@@ -0,0 +1,18 @@
+import React, { useContext, useEffect } from 'react';
+import * as ActualSentryFromSDK from '@sentry/browser';
+import { CookieConsentContext } from './CookieConsent';
+
+export default function Sentry() {
+  const cookiesConsented = useContext(CookieConsentContext);
```
Maurice22, line 6: Small detail, but for consistency it would be nice to either stick to cookiesConsented or cookiesAccepted

src/util/CookieConsent.jsx
```diff
@@ -0,0 +1,42 @@
...
+    <div className="fixed left-0 bottom-0 w-full bg-gray-900 p-4 text-white font-exo2 z-50 flex flex-no-wrap items-center space-between flex-row">
+      <div className="flex-grow">{t('App.usesCookies')}</div>
+      <button
+        type="button"
+        className="bg-primary p-2 rounded ml-2 font-open-sans flex items-center justify-center"
+        onClick={onClick}
+      >
+        Okay
+      </button>
+    </div>
```
Maurice22, line 18: scroll up button has a higher z-index -> I would vote for using the `z-` class on the scroll up button and the report buttons (instead of setting the z-index in the css) to make this streamlined

src/util/CookieConsent.jsx
```diff
@@ -0,0 +1,42 @@
...
+  const { t } = useTranslation();
+  return (
+    <div className="fixed left-0 bottom-0 w-full bg-gray-900 p-4 text-white font-exo2 z-50 flex flex-no-wrap items-center space-between flex-row">
+      <div className="flex-grow">{t('App.usesCookies')}</div>
+      <button
+        type="button"
+        className="bg-primary p-2 rounded ml-2 font-open-sans flex items-center justify-center"
+        onClick={onClick}
+      >
+        Okay
```
Maurice22, line 16: I would rename this button to sth. "stronger", i.e. "I accept"/"Einverstanden". Maybe we should also add a disagree button

src/util/CookieConsent.jsx
```diff
@@ -0,0 +1,42 @@
...
+  const acceptCookies = () => {
+    window.localStorage.setItem('qh-cookies-accepted', 'true');
+    setCookiesAccepted(true);
+  };
+
+  useEffect(() => {
+    const value = window.localStorage.getItem('qh-cookies-accepted');
+    const isAccepted = value === 'true';
+    setCookiesAccepted(isAccepted);
+  }, [cookiesAccepted]);
```
Maurice22, line 34: Why are we doing this here? As far as I can tell you are setting the cookiesAccepted above and run this again afterwards. It should come to the same result as above.
I guess what I mean is: Shouldn't this be only executed on first render and not whenever the cookieAccept state changes

src/App.jsx
```diff
@@ -1,304 +1,27 @@
...
-        <Route path="*">
-          <Page>
-            <NotFound />
-          </Page>
-        </Route>
-      </Switch>
-    </Router>
+    <Suspense fallback={<Loader />}>
+      <CookieConsent>
+        <Sentry />
```
Maurice22, line 20: I don't know, somehow this feels wrong. Shouldn't we use sth. like useScrollToTop() in the JavaScript code instead of having those components in the "HTML"? Especially for the Sentry Component

src/components/Routes.jsx
```diff
@@ -0,0 +1,152 @@
...
+import NotifyMe from '../views/NotifyMe';
+import CompleteOfferHelp from '../views/CompleteOfferHelp';
+import Main from '../views/Main';
+import NotFound from '../views/NotFound';
+import Page from './Page';
+import { CookieConsentContext } from '../util/CookieConsent';
+
+
+function usePageViews() {
+  const cookiesConsented = useContext(CookieConsentContext);
```
Maurice22, line 30: Same naming remark as below

src/App.jsx
florianschmidt1994: Yeah we should, but that would mean we have to introduce yet another wrapper component that we can use the execute the hook, as hooks can't be placed within jsx. For the sake of simplicity I've kept it this way. Maybe you / someone knows another way to do this?

Maurice22: We could move sentry to the Routes component

src/util/Sentry.jsx
florianschmidt1994: I've went with cookiesConsented everywhere

src/util/CookieConsent.jsx
florianschmidt1994: I've removed the dependency

src/App.jsx
florianschmidt1994: Yeah, but that would again be a weird entanglement of concerns, right?

Maurice22:
While we are at it, should we move the ScrollToTop component from /components to /utils as well?

src/util/CookieConsent.jsx
florianschmidt1994: @tgraupne what's your take on this?

src/util/CookieConsent.jsx
florianschmidt1994: I've used inline styles for this for now. I've re-opened #213

src/util/CookieConsent.jsx
```diff
@@ -6,7 +6,10 @@ export const CookieConsentContext = React.createContext(false);
 function CookieConsentBanner({ onClick }) {
   const { t } = useTranslation();
   return (
-    <div className="fixed left-0 bottom-0 w-full bg-gray-900 p-4 text-white font-exo2 z-50 flex flex-no-wrap items-center space-between flex-row">
+    <div
+      style={{ zIndex: 501 }}
+      className="fixed left-0 bottom-0 w-full bg-gray-900 p-4 text-white font-exo2 z-50 flex flex-no-wrap items-center space-between flex-row"
```
Maurice22, line 11: I guess you can get rid of z-50 then

src/util/CookieConsent.jsx
kenodressel: I agree with @Maurice22 on this

src/components/Routes.jsx
```diff
@@ -0,0 +1,152 @@
...
+  const location = useLocation();
+
+  useEffect(() => {
+    if (cookiesConsented) {
+      firebase.analytics().logEvent('page_view', { page_path: location.pathname });
```
kenodressel, line 35: Looks to me like we init a new instance of firebase analytics every time we change the page. Not sure if this is intended. Previously we had a single instance that was reused. There are some events that get sent on initialization: https://support.google.com/firebase/answer/6317485

src/index.jsx
```diff
@@ -10,17 +10,10 @@ import './styles/tailwind.css';
 import './styles/index.css';
 import './styles/App.css';
 import App from './App';
-import Loader from './components/Loader';
 import * as serviceWorker from './serviceWorker';
 import './i18n'; // Internalization
 
-ReactDOM.render(
-  <Suspense fallback={<Loader />}>
-    {/* Suspense is necessary for internalization */}
```
kenodressel, line 19: Could we keep this comment?

florianschmidt1994: There is a comment in the App.jsx about why we use suspense. It is not in the JSX but above in the code, I've moved it there for readability

src/util/CookieConsent.jsx
florianschmidt1994: About which part? renaming to "I accept / Einverstanden" with i18n should be easy to do, but having a disagree button might logically collide with firebase writing to indexedDB, i.e. for session data

Maurice22: Many websites show cookie popus where you can't deselect the "necessary" cookies but you can deselect the analytics cookies. Maybe we can add buttons like "Continue with all cookies" and "continue with necessary cookies"?
@tgraupne?

src/App.jsx
```diff
@@ -1,304 +1,27 @@
...
-import useScrollToTop from './components/ScrollToTop';
-import ShareButtons from './components/ShareButtons';
-import Press from './views/Press';
-import createEventListener from './util/createEventListener';
-import Security from './views/Security';
-import arrowUpIcon from './assets/arrows_up.svg';
-import FilteredListFrame from './views/FilteredListFrame';
-import Footer from './components/Footer';
+import Routes from './components/Routes';
+import Loader from './components/Loader';
```
Maurice22, line 5: @florianschmidt1994, when you resolve the merge conflict, you should change this to ./components/loader/Loader

src/components/Routes.jsx
florianschmidt1994: > `firebase.analytics()` just returns a singleton object, the same every time. All of the Firebase product entrypoints exposed by the firebase object are all that way. Whatever method you want to use to get that singleton object is completely up to you.

From https://stackoverflow.com/a/59934906/4187312

tgraupne:
Hi @florianschmidt1994, what would you think about splitting your changes into multiple feature/pull requests? I think this pr has surpassed the issue of refactoring the cookie banner by far :-D

florianschmidt1994:
@tgraupne I would like to leave it this way, I believe most of the changes are actually related to the cookie banner

kenodressel:
So where are we on this @florianschmidt1994 @Maurice22?

Maurice22:
AFAIK @tgraupne is having a look into how such a cookie banner has to look like to be compliant with the regulations. Afterwards I guess @florianschmidt1994 will have to adapt these changes in the PR.

florianschmidt1994:
@mauriceackel could you re-review this considering that we agreed to move forward with this structural refactoring regardless of the GDPR implications?

PR #231, opened 2020-04-16
Catch errors when parsing entries

### LINKED ISSUE #220: Displaying the map causes crash in dev environment
When clicking on the button to display the map view it causes a crash after quickly displaying the map. 

Browser: Safari 13.0.5
OS: MacOS
Logged In / Out: Doesn't matter
Address selected yes / no: Doesn't matter

```
Unhandled Rejection (FirebaseError): Function GeoPoint() requires its first argument to be of type number, but it was: undefined
```

```jsx
src/components/EntryMap.jsx:39
  36 | const fetchEntries = async () => {
  37 |   const queryResult = await fb.store.collection('ask-for-help').get();
  38 | 
> 39 |   const entriesFromQuery = queryResult.docs
  40 |     .map((document) => ({ ...document.data().d, id: document.id }))
  41 |     .map((dataPoint) => ({
  42 |       type: 'Feature',
```

florianschmidt1994:
Some entries have lat/lon zero which leads to an error
when we try to extract the data with the firebase library.

Instead of crashing when one entry has this problem, we
are now filtering out those invalid entries and logging
an error instead.

Note that in production we should not have those entries
and they indicate that something is wrong in another place,
which is why I've added reporting to sentry if any of those
invalid entries should occur.

Close #220

### DIFF
--- a/src/components/EntryMap.jsx
+++ b/src/components/EntryMap.jsx
@@ -3,7 +3,7 @@ import GoogleMapReact from 'google-map-react';
 import '../styles/Map.css';
 import useSupercluster from 'use-supercluster';
 import { useTranslation } from 'react-i18next';
-
+import * as Sentry from '@sentry/browser';
 import fb from '../firebase';
 import Entry from './entry/Entry';
 import NotifyMe from './NotifyMe';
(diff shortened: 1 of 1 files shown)


src/components/EntryMap.jsx
```diff
@@ -32,12 +32,24 @@ export default function EntryMap() {
 
   const [entries, setEntries] = useState([]);
 
+  function parseDoc(doc) {
```
Maurice22, line 35: I think we defined our functions like this in the past: `const parseDoc = (doc) => {};`. Don't know if it is important to stick to this

src/components/EntryMap.jsx
```diff
@@ -32,12 +32,24 @@ export default function EntryMap() {
...
+    try {
+      return { ...doc.data().d, id: doc.id };
+    } catch (err) {
+      Sentry.captureException(new Error(`Error parsing ask-for-help ${doc.id}`));
+      // eslint-disable-next-line no-console
+      console.error(`Error parsing ask-for-help ${doc.id}`);
```
Maurice22, line 41: Shouldn't we prevent writing errors to the console as suggested by the linting rule? I mean it does not add a benefit to the user and the error is caught in sentry anyways.

florianschmidt1994: Yeah good point. I'll remove it

src/components/EntryMap.jsx
florianschmidt1994: Hm I'll change it for the sake of consistency, but it doesn't make a difference here

Maurice22:
@florianschmidt1994 Don't we need the same functionality in EntryList.jsx?

PR #233, opened 2020-04-16
Feature/127 address clickable

### LINKED ISSUE #127: Feature: make address on post clickable and use as filter
Idea: Make the address on a post on the overview page clickable and use it to filter the existing entries. Could be a cool feature to quickly filter for entries in a specific location, without having to navigate back to the main filter

Screenshot 2020-03-23 at 14 58 20

e.g. when clicking on  **30167 Hannover, Deutschland**, it would show me all posts for that location (probably resetting the entered radius then)

Maurice22:

- Makes the address inside an Entry clickable.
- If entry is displayed in the entry list, we will set a new search filter.
- If the entry is displayed in another component, we redirect to the overview page and pass a query param.
- The entry list now parses the query params on load and if an `address` param is set, we set this as initial filter.
- **Fixed an issue where searching for an address that does not exist simply displayed all entries. Instead, we will now show an empty list.**

Closes #127


Maurice22:
No as the entry component does not know where it is rendered. If it is rendered in the EntryList, we should not change the location imo.

src/components/EntryList.jsx
```diff
@@ -118,7 +118,15 @@ export default function EntryList({ pageSize = 0 }) {
   };
 
   useEffect(() => {
-    loadDocuments(buildQuery());
+    const urlParams = new URLSearchParams(window.location.hash.split('?')[1]);
```
Maurice22, line 121:
```suggestion
    // Because we use the hash router, we cannot use the default functionality here as it would expect the query parameters before the hash
    const urlParams = new URLSearchParams(window.location.hash.split('?')[1]);
```

src/components/entry/Entry.jsx
```diff
@@ -113,6 +114,16 @@ export default function Entry(props) {
     setPopupVisible(false);
   };
 
+  const handleAddressClick = (e) => {
+    // Only handle if handler is registered
+    if (onAddressClick) {
+      e.preventDefault();
+      e.stopPropagation();
+
```
jagaSto, line 122: unnecessary empty line :)

src/components/entry/Entry.jsx
```diff
@@ -113,6 +114,16 @@ export default function Entry(props) {
...
+      onAddressClick(location);
```
jagaSto, line 123:
```suggestion
      onAddressClick(location);
```

Maurice22:
Decided to always have this functionality activated, specify a default behaviour inside the entry and only allow an override if necessary

Maurice22:
I streamlined the whole functionality so that now the redirect is triggered inside the entry. Hence, this behavior is always on. If a component wants to overwrite the default behavior (i.e. redirect to /overview) it can specify an onAddressClick handler explicitly.

src/components/EntryList.jsx
```diff
@@ -118,8 +120,17 @@ export default function EntryList({ pageSize = 0 }) {
   };
 
   useEffect(() => {
-    loadDocuments(buildQuery());
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+    // Because we use the hash router, we cannot use the default functionality here as it would expect the query parameters before the hash
+    const urlParams = new URLSearchParams(window.location.hash.split('?')[1]);
```
jagaSto, line 124: we could use `windowLocation` here now instead of window.location, right?

src/components/entry/Entry.jsx
```diff
@@ -114,6 +115,19 @@ export default function Entry(props) {
...
+    e.stopPropagation();
+    if (onAddressClick) {
+      onAddressClick(location);
+    } else {
+      history.push({
+        pathname: '/overview',
+        search: `?address=${location}`,
+      });
+    }
+  };
```
jagaSto, line 129: nice, much easier to comprehend now!

src/components/EntryList.jsx
Maurice22: Not quite but I changed it to work like this

tgraupne:
Hi @Maurice22, why does this "need consideration"?
Is there something I am missing here :-)

Maurice22:
Once upon a time ;) we labeled our PRs this way to signal whether it is a straight forward change or whether the review should be done more thoroughly. This is such a PR where the review should be done in more detail.

florianschmidt1994:
I've tried and tested this. Works well for me! Thanks for implementing this @Maurice22 Will now merge!

Issue #218, opened 2020-04-09
Enhance emails with "report post" feature


jagaSto:
We enhance the emails we currently sent out to users in which we inform them about new help postings in their area with the possibility to report the post as malicious, or add a hint that indicates that it is possible to report said entry as malicious on the website (users suggested this before)

Maurice22:
We merged PR #236 as a precondition for this information to be included in the email. @florianschmidt1994 will update the email template soon to add a text along the lines of "Wenn Du glaubst, dass dieses Inserat unangemessene Inhalte enthält, kannst du es <hier>(use link from email template) melden."

jagaSto:
sendigrid:

```
<div style="font-family: inherit"><br></div>
<div style="font-family: inherit; line-height: 1.5">Wenn Du glaubst, dass dieses Inserat unangemessene Inhalte enthält, kannst du es <a href="{{{reportLink}}}">hier</a> melden.</div>
<div style="font-family: inherit"><br></div>
```

jagaSto:
<img width="643" alt="Screenshot 2020-06-27 at 13 24 45" src="https://user-images.githubusercontent.com/20641332/85921068-92b35000-b879-11ea-9945-6949a95c2f1a.png">

jagaSto:
done!

PR #236, opened 2020-04-19
Query param to display entry in attempt report state

Maurice22:

As suggested in #218, adding a link to report an entry would be nice. This PR introduces a query param "reported" to the /offer-help/{id} view. When set, the entry on this page will be displayed in the attemptReport state so that a user can report the post there. Furthermore, it passes this url to the sendgrid template.

Before this can be merged, we need to adapt the sendgrid template.

### DIFF
--- a/firebase/functions/index.js
+++ b/firebase/functions/index.js
@@ -157,10 +157,11 @@ async function searchAndSendNotificationEmails() {
           from: 'help@quarantaenehelden.org',
           templateId: 'd-9e0d0ec8eda04c9a98e6cb1edffdac71',
           dynamic_template_data: {
-            subject: 'QuarantäneHelden - Jemand braucht deine Hilfe!',
+            subject: 'QuarantäneHelden - Jemand braucht Deine Hilfe!',
             request: askForHelpSnapData.d.request,
             location: askForHelpSnapData.d.location,
             link: `https://www.quarantaenehelden.org/#/offer-help/${askForHelpId}`,
+            reportLink: `https://www.quarantaenehelden.org/#/offer-help/${askForHelpId}?report=true`,
           },
           hideWarnings: true, // removes triple bracket warning
         });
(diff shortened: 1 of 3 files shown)


src/views/OfferHelp.jsx
```diff
@@ -56,15 +57,19 @@ export default function OfferHelp() {
     return history.push('/success-offer');
   };
 
-  useEffect(getUserData, []);
+  useEffect(() => {
+    getUserData();
+    const urlParams = new URLSearchParams(window.location.hash.split('?')[1]);
+    setReport(!!urlParams.get('report'));
```
jagaSto, line 63: Screenshot 2020-04-20 at 19 46 00

This will always return true for the string `false`.

You could change this e.g. to `setReport(urlParams.get('report').toLowerCase() === 'true');` Not really pretty, but works.

Maurice22: I disagree to check for a certain value, as i.e. in javascript also 1 evaluates to true and 0 to false. If the element should not be reported, then the query parameter should not be set. I changed the code to only check for existence now instead of casting the value to boolean. This also enables /foobar?report instead of having to specify a value.

firebase/functions/index.js
```diff
@@ -157,10 +157,11 @@ async function searchAndSendNotificationEmails() {
           from: 'help@quarantaenehelden.org',
           templateId: 'd-9e0d0ec8eda04c9a98e6cb1edffdac71',
           dynamic_template_data: {
-            subject: 'QuarantäneHelden - Jemand braucht deine Hilfe!',
+            subject: 'QuarantäneHelden - Jemand braucht Deine Hilfe!',
             request: askForHelpSnapData.d.request,
             location: askForHelpSnapData.d.location,
             link: `https://www.quarantaenehelden.org/#/offer-help/${askForHelpId}`,
+            reportLink: `https://www.quarantaenehelden.org/#/offer-help/${askForHelpId}?report=true`,
```
jagaSto, line 164:
```suggestion
            reportLink: `https://www.quarantaenehelden.org/#/offer-help/${askForHelpId}?report`,
```

PR #242, opened 2020-04-21
add redirect url for sendEmailVerification

jagaSto:

* adds redirect URL for email verification, which allows the user to navigate back to the website
* does **not** automatically redirect the user to the website (this is not possible with the selected auth method, as far as my research suggested)

Screenshot 2020-04-21 at 23 11 04

### DIFF
--- a/src/views/Signup.jsx
+++ b/src/views/Signup.jsx
@@ -41,7 +41,7 @@ export default () => {
 
     try {
       const signUpResult = await createUserWithEmailAndPassword(email, password);
-      await signUpResult.user.sendEmailVerification({ url: `${baseUrl}/${returnUrl}` });
+      await signUpResult.user.sendEmailVerification({ url: `${baseUrl}/#/${returnUrl}` });
     } catch (err) {
       switch (err.code) {
         case 'auth/email-already-in-use': setError(t('views.signUp.emailInUse')); break;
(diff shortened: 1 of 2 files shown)


Maurice22:
Should we think about implementing our own frontend page that handles all the actions as explained here:
https://firebase.google.com/docs/auth/custom-email-handler

src/views/Signup.jsx
```diff
@@ -40,7 +41,7 @@ export default () => {
 
     try {
       const signUpResult = await createUserWithEmailAndPassword(email, password);
-      await signUpResult.user.sendEmailVerification();
+      await signUpResult.user.sendEmailVerification({ url: `${baseUrl}/${returnUrl}` });
```
Maurice22, line 44: This is lacking the hash sign and hence does not redirect to the right location. Should be like in my suggestion.
```suggestion
      await signUpResult.user.sendEmailVerification({ url: `${baseUrl}/#/${returnUrl}` });
```

src/views/VerifyEmail.jsx
```diff
@@ -19,7 +20,7 @@ export default () => {
 
   const resendEmail = async (e) => {
     e.preventDefault();
-    await user.sendEmailVerification();
+    await user.sendEmailVerification({ url: `${baseUrl}/ask-for-help` });
```
Maurice22, line 23: Same as above.
```suggestion
    await user.sendEmailVerification({ url: `${baseUrl}/#/ask-for-help` });
```

jagaSto:
We can definitely think about it since I don't see a way to perform an automatic redirect with the current implementation! Nevertheless, I would still merge this, since it will probably take a bit until we have discussed and implemented the proposition :)

PR #243, opened 2020-04-22
Fixed data retrieval for geofirestore

Maurice22:

Depending on which way we use to retrieve data (i.e. firestore or geofirestore) we need to access a different key. Was broken in last master in last deploy.


florianschmidt1994:
Merging

PR #244, opened 2020-04-23
Remove QH logo duplication on mobile main page

### LINKED ISSUE #235: Remove logo duplication on frontpage

jakob-ed:

From the `Main` component, removes a logo which is only shown on mobile so that only the logo from the `MobileTopNavigation` component is visible. Closes #235.


florianschmidt1994:
@tgraupne no we should be fine!

@jakob-ed Thanks for fixing this. Changes look to to me as well :)

PR #248, opened 2020-04-24
Feature/headless cms

mauriceackel:

Added a headless CMS, powered by Drive and Firebase Realtime Database.

* Loads content for Security Page, FAQ and Press dynamically
* Added markdown rendering engine for markdown support
* Refactored Loader to wait on a condition
* Moved firebase download link creation into own utility

**IMPORTANT: DO NOT MERGE BEFORE THE CONTENT IS UPDATED IN THE CMS**

**TODOs before merging**
- [x] Create service account on prod for publishing
- [x] Connect CMS backend to prod and publish
- [x] Set prod security rules for realtime DB (read collections defined by CMS, no write) 

```
"cmsContent": {
  ".read": true
}
```

**Others details**
- [x] This PR introduces a new dependency, "markdown-to-jsx". Reason: Being able to define content in markdown.


src/util/StyledMardown.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import Markdown from 'markdown-to-jsx';
+
+export default function StyledMardown(props) {
```
jagaSto, line 4: typo, should be `StyledMarkdown`

src/util/useCms.jsx
```diff
@@ -0,0 +1,38 @@
...
+          }
+          return resolve(contentRef);
+        });
+      })
+    );
+
+    const getInternationalizedContent = async () => {
+      for (let i = 0; i < languages.length; i += 1) {
+        try {
+          // eslint-disable-next-line no-await-in-loop
```
jagaSto, line 26: you could use a [for ... of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) loop here instead, where async operations are allowed :)

src/util/useCms.jsx
```diff
@@ -0,0 +1,38 @@
...
+          const contentRef = await getContent(collectionName, languages[i]);
+          contentRef.on('value', (snapshot) => setContent(snapshot.val()));
+          return;
+        } catch (err) { } // eslint-disable-line no-empty
```
jagaSto, line 30: should we add sentry logging for the error case?

src/util/useFirebaseDownload.jsx
```diff
@@ -0,0 +1,22 @@
...
+import * as firebase from 'firebase/app';
+
+export default function useFirebaseDownload(url) {
+  const [link, setLink] = useState('');
+  const [error, setError] = useState(null);
+
+  useEffect(() => {
+    const getLink = async () => {
+      try {
+        const l = await firebase.storage().refFromURL(url).getDownloadURL();
```
jagaSto, line 11: could we rename this variable into something more explanatory, e.g. `linkFromFirebase` (since `link` and `url` are already taken)? 😊

src/util/useCms.jsx
Maurice22: No, unfortunately I can't as this id forbidden by ESLint

src/util/useCms.jsx
Maurice22: No, as I only use this error so that I can iterate over all languages identified by i18next (i.e. [en-US, en, de]). If we don't find the first language in the cms, we throw the error to continue to look for the next language. The only think that would make sense is to report the error if there was no match for all tested languages.

Maurice22: I implemented my approach, can you check on it?

mauriceackel:
@kenodressel, @florianschmidt1994, @jagaSto I fixed the merge conflict, can someone have a look?

kenodressel:
Current state:  we are having issues with the security as a malicious actor could just delete the source files from gdrive after defacing the website and there is no easy way back from this. Therefore this PR is blocked.

mauriceackel:
Just tested on prod. Works fine.

PR #252, opened 2020-04-26
add analytics events for engagement with entry

jagaSto:
 
This PR introduces three new analytics events for the engagement with `ask-for-help` entries, namely:
* `help_request_deleted`
* `help_request_solved`
* `help_request_reported`

Adding these events allow us to run better analytics in the future. I tried to align the naming with the existing events, e.g. [success_request_help](https://github.com/kenodressel/quarantine-hero/blob/21075a4e8c50898dab9e3bf39617b767579914a3/src/views/Success.jsx#L8)

### DIFF
--- a/src/components/entry/Entry.jsx
+++ b/src/components/entry/Entry.jsx
@@ -69,6 +69,7 @@ export default function Entry(props) {
     await fb.store.collection('/deleted').doc(id).set({
       collectionName, ...doc.data(),
     });
+    await fb.analytics.logEvent('help_request_deleted');
     setDeleted(true);
     setAttemptingToDelete(false);
     setPopupVisible(true); // trigger the deletion confirmation popup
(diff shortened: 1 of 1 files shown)


src/components/entry/Entry.jsx
```diff
@@ -135,6 +137,7 @@ export default function Entry(props) {
       timestamp: Date.now(),
     };
     await reportedPostsCollection.add(data);
+    await fb.analytics.logEvent('help_request_reported');
```
Maurice22, line 140: Why do we need to await those?

kenodressel:
We should check what happens when analytics is undefined

src/components/entry/Entry.jsx
jagaSto: thought they would be async. nvm

Maurice22:
@jagaSto LGTM. Other than that, can you please check the comment from Keno regarding i.e. the case in which cookies are not consented.

Maurice22:
@kenodressel I just checked the code and because we initialize the fb.analytics event with a dummy function () => {} I think it should not become undefined. Because the only other thing we do with it is assigning it the real firebase analytics object once the cookies are consented. What do you think?

kenodressel:
Sounds good to me.

PR #262, opened 2020-04-27
Add askForHelpId to sendgrid dynamic template variables for offer-help

### LINKED ISSUE #261: Add deep-link to entry in email
~As part of our phone hotline we should add a link back to the entry to the emails. Not only will this allow hotline members to easily detect answers to entries, but it will also be helpful to show users that they can mark their entry as solved~

EDIT to make the scenario a little more clear:

A person (helpee) posts an entry (entry) to our website asking for help. Another person (helper) offers to help them. (offer-help)
We already send an email to the helpee that somebody is offering help. This email does currently not include a link to the entry. With the PR solving this issue we will be able to add such a link to the email

Related to #245, possible duplicate of #245

florianschmidt1994:
When a person asks for help (helpee) receives an offer from someone (helper) to help, we send an email to the helpee. Within this email we want to add a link back to the place where entry which was created by the helpee when asking for help. 

This PR adds the id of this entry in order to modify the email template to do this.

Closes #261

### DIFF
--- a/firebase/functions/index.js
+++ b/firebase/functions/index.js
@@ -39,35 +39,27 @@ async function onOfferHelpCreate(snap) {
     const { email: receiver } = data.toJSON();
     const { answer, email } = offer.data();
 
-    // eslint-disable-next-line no-console
-    console.log({
+    const sendgridOptions = {
       to: receiver,
-      from: email,
+      from: 'help@quarantaenehelden.org',
       templateId: 'd-ed9746e4ff064676b7df121c81037fab',
+      replyTo: { email },
+      hideWarnings: true, // removes triple bracket warning
       dynamic_template_data: {
         subject: 'QuarantäneHeld*innen - Jemand hat Dir geschrieben!',
         answer,
         email,
         request,
+        askForHelpId: askForHelp.id,
       },
-    });
+    };
+
+    // eslint-disable-next-line no-console
+    console.log(sendgridOptions);
+
     try {
       if (SEND_EMAILS) {
-        await sgMail.send({
-          to: receiver,
-          from: 'help@quarantaenehelden.org',
-          replyTo: {
-            email,
-          },
-          templateId: 'd-ed9746e4ff064676b7df121c81037fab',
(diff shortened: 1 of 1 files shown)


firebase/functions/index.js
```diff
@@ -39,35 +39,27 @@ async function onOfferHelpCreate(snap) {
...
+      from: 'help@quarantaenehelden.org',
       templateId: 'd-ed9746e4ff064676b7df121c81037fab',
+      replyTo: { email },
+      hideWarnings: true, // removes triple bracket warning
       dynamic_template_data: {
         subject: 'QuarantäneHelden - Jemand hat dir geschrieben!',
         answer,
         email,
         request,
+        askForHelpId: askForHelp.id,
```
Maurice22, line 53: Should we use the complete link here? We do so as well in other places. I think we should not have the base URL in the sendgrid template in case it needs to be changed at some point.

Maurice22:
@florianschmidt1994 Regarding the merge conflict. It is due to the change of the subject line so be aware of using the updated one when resolving the conflict.

florianschmidt1994:
@Maurice22 I've updated this

firebase/functions/index.js
```diff
@@ -39,35 +39,27 @@ async function onOfferHelpCreate(snap) {
...
+      from: 'help@quarantaenehelden.org',
       templateId: 'd-ed9746e4ff064676b7df121c81037fab',
+      replyTo: { email },
+      hideWarnings: true, // removes triple bracket warning
       dynamic_template_data: {
         subject: 'QuarantäneHeld*innen - Jemand hat Dir geschrieben!',
         answer,
         email,
         request,
+        askForHelpLink: `https://www.quarantaenehelden.org/#/offer-help/${askForHelp.id}`,
```
Maurice22, line 53: Doesn't it need to be `askRecord.id` here as askForHelp is only a ref?

florianschmidt1994: No, this works as intended. I just verified it on staging again :)

PR #270, opened 2020-04-30
Own account management

### LINKED ISSUE #241: Improve email-verification redirect logic
Currently, when attempting to verify your email address, you need to click the verification link in the email, which takes you to the google auth success page. It would be better UX however if the link would redirect you to the website again after successful verification, similar to the sign-in logic for the `notify-me` component.
Also, if you verify your email address and do *not* reload the page, you need to re-login for the website to notice that your email is indeed verified. This is also bad UX and could be resolved as well by improving the redirect logic to automatically redirect the user, and reload the page automatically.

Maurice22:

* Adds a new view which handles email action URLs generated by firebase
* Depending on the opcode, either email verification, password reset or email recovery is handled
* Closes #241 

**IMPORTANT: Once we merge this, we have to update the firebase settings under Authentication -> Email Templates and set the actionURL to 'https://quarantaenehelden.org/#/handle-email-action'** 

**Others details**
For testing the password recovery functionality, you have to change your account password. You can do so by adding the following code to a dummy component and clicking the button:
```
<button
  type="button"
  onClick={() => {
    user.updateEmail('test@example.com');
  }}
>
  TestEmail change
</button>
```

### DIFF
--- a/public/locales/de/translation.json
+++ b/public/locales/de/translation.json
@@ -246,8 +246,8 @@
       "resendSuccess": "Eine E-Mail mit Bestätigungs-Link wurde Dir zugesendet!"
     },
     "emailActions": {
-      "invalidToken": "Diese Aktion ist leider nicht mehr verfügbar, da Dein Token abgelaufen ist. Bitte versuche es erneut.",
-      "unknownAction": "Leider kann die gewünschte Aktion nicht ausgeführt werden, da sie uns unbekannt ist.",
+      "invalidToken": "Diese Aktion ist leider nicht mehr verfügbar, da Dein Link abgelaufen ist. Bitte versuche es erneut.",
+      "unknownAction": "Leider kann die gewünschte Aktion nicht ausgeführt werden, da sie uns unbekannt ist. Bitte versuche es erneut. Falls der Fehler weiter besteht, kannst Du Dich gerne unter help@quarantaenehelden.org bei uns melden.",
       "continue": "Weiter",
       "signIn": {
         "invalidEmail": "Die angegebene E-Mail-Adresse ist ungültig. Bitte versuche es erneut",
--- a/public/locales/de/translation.json
+++ b/public/locales/de/translation.json
@@ -72,7 +72,6 @@
       }
     },
     "completeOfferHelp": {
-      "mailYouRegistered": "Bitte gib Deine E-Mail-Adresse ein, mit der Du Dich registriert hast",
       "youAreHero": "Du bist eine Held*in!",
       "mailVerified": "Deine E-Mail-Adresse wurde verifiziert! Du wirst nun von uns per E-Mail benachrichtigt werden, wenn jemand in",
       "needsHelp": "Hilfe benötigt",
@@ -250,6 +249,13 @@
       "invalidToken": "Diese Aktion ist leider nicht mehr verfügbar, da dein Token abgelaufen ist. Bitte versuche es erneut.",
       "unknownAction": "Leider kann die gewünschte Aktion nicht ausgeführt werden, da sie uns unbekannt ist.",
       "continue": "Weiter",
+      "signIn": {
+        "invalidEmail": "Die angegebene E-Mail-Adresse ist ungültig. Bitte versuche es erneut",
+        "invalidLink": "Die Aktion kann leider nich durgeführt werden, da der benutze Link ungülig ist",
+        "mailYouRegistered": "Bitte gib die E-Mail-Adresse ein, mit der Du Dich registriert hast",
+        "yourEmail": "Deine E-Mail-Adresse",
+        "email": "E-Mail-Adresse"
+      },
       "resetPassword": {
         "heading": "Hier kannst Du Dein Passwort zurücksetzen",
         "newPassword": "Neues Passwort",
--- a/public/locales/de/translation.json
+++ b/public/locales/de/translation.json
@@ -249,6 +249,7 @@
     "emailActions": {
       "invalidToken": "Diese Aktion ist leider nicht mehr verfügbar, da dein Token abgelaufen ist. Bitte versuche es erneut.",
       "unknownAction": "Leider kann die gewünschte Aktion nicht ausgeführt werden, da sie uns unbekannt ist.",
+      "continue": "Weiter",
       "resetPassword": {
         "heading": "Hier kannst Du Dein Passwort zurücksetzen",
         "newPassword": "Neues Passwort",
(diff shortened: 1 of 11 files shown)


public/locales/de/translation.json
```diff
@@ -238,9 +237,52 @@
...
       "ItBusy": "Unsere IT-Held*innen sind leider gerade beschäftigt, kümmern sich aber um das Problem, sobald Kapazitäten frei werden.",
       "toHome": "Bring mich zur Startseite"
+    },
+    "emailActions": {
+      "invalidToken": "Diese Aktion ist leider nicht mehr verfügbar, da dein Token abgelaufen ist. Bitte versuche es erneut.",
+      "unknownAction": "Leider kann die gewünschte Aktion nicht ausgeführt werden, da sie uns unbekannt ist.",
+      "continue": "Weiter",
+      "signIn": {
+        "invalidEmail": "Die angegebene E-Mail-Adresse ist ungültig. Bitte versuche es erneut",
+        "invalidLink": "Die Aktion kann leider nich durgeführt werden, da der benutze Link ungülig ist",
```
florianschmidt1994, line 247: ```"invalidLink": "Die Aktion kann leider nich durchgeführt werden, da der benutze Link ungülig ist",```

src/views/CompleteOfferHelp.jsx
```diff
@@ -35,7 +41,11 @@ export default function CompleteOfferHelp() {
       lng = coordinates.lng;
     }
 
-    await offerHelpCollection.add({
+    const idHashBuffer = await crypto.subtle.digest('SHA-1', new TextEncoder().encode(`${email}_${plz}`));
+    const idHashArray = Array.from(new Uint8Array(idHashBuffer)); // convert buffer to byte array
+    const idHash = idHashArray.map((b) => b.toString(16).padStart(2, '0')).join('');
```
florianschmidt1994, line 46: Can we move all the hashing into a utility function and give it nice name?

src/views/CompleteOfferHelp.jsx
```diff
@@ -48,78 +58,57 @@ export default function CompleteOfferHelp() {
...
-            {t('views.completeOfferHelp.needsHelp')}
-          </p>
-          <div className="flex justify-center flex-col items-center mb-8">
-            <img className="h-48 w-48 my-10" src={require('../assets/success.svg')} alt="" />
-            <Link className="btn-green mt-10" to="/dashboard">{t('views.completeOfferHelp.toYourOverview')}</Link>
-          </div>
-        </>
-      )}
-    </div>
+    <Loader waitOn={!isLoading}>
```
florianschmidt1994, line 81: I like the idea! The naming is a little confusing to me with `waitOn`, what do you think about just using `loading={isLoading}` or something similar?

src/views/CompleteOfferHelp.jsx
```diff
@@ -48,78 +58,57 @@ export default function CompleteOfferHelp() {
...
+      <div className="p-4">
+        {error
```
florianschmidt1994, line 83: My personal preference is to use two return blocks instead of inline js in jsx, i.e. ```js
if (condition) return <></> else return <></>``` Any opinions on that? Feel free to keep it that way if you like it better

src/views/CompleteOfferHelp.jsx
Maurice22: You are right, the naming is a bit confusing. However, I kinda already used the same implementation my other PRs, so I'm going to create a follow up issue for me to streamline it once everything is merged in order to not overwrite it in future PRs.

src/views/CompleteOfferHelp.jsx
Maurice22: I agree, changed it 👍

src/views/CompleteOfferHelp.jsx
Maurice22: Moved to own utility function

Maurice22:
Sorry for that! Should be fixed now. The issue was that I used the wrong variable name after a small refactoring.

src/views/HandleEmailAction.jsx
```diff
@@ -0,0 +1,446 @@
...
+    const paramArray = [...new URLSearchParams(location.search).entries(), ...new URLSearchParams(window.location.search)];
+    for (let i = 0; i < paramArray.length; i += 1) {
+      const [key, value] = paramArray[i];
+      paramObj[key] = value;
+    }
+
+    // Firebase requires the continue URL to have a domain part. As we use Links for routing, we strip the domain part off.
+    paramObj.continueUrl = paramObj.continueUrl && paramObj.continueUrl.split('#')[1];
+    setUrlParams(paramObj);
+  }, [location]);
```
florianschmidt1994, line 30: We could move this whole section into a hook itself and use it like `const urlParams = useUrlParams();` for readability

florianschmidt1994:
It works now for me. I've added one more remark above, otherwise this is looking good

src/views/HandleEmailAction.jsx
Maurice22: This is actually sth. that @jagaSto has implemented in his recent PR. I would say I then simply wait for his PR to be merged and use the hook here then.

public/locales/de/translation.json
```diff
@@ -245,9 +244,52 @@
       "recheckEmail": "Bestätigung überprüfen",
       "resendEmail": "E-Mail erneut senden",
       "resendSuccess": "Eine E-Mail mit Bestätigungs-Link wurde Dir zugesendet!"
+    },
+    "emailActions": {
+      "invalidToken": "Diese Aktion ist leider nicht mehr verfügbar, da Dein Token abgelaufen ist. Bitte versuche es erneut.",
```
kenodressel, line 249: Could we make this a bit more user friendly? I don't think people know what a "token" is.

public/locales/de/translation.json
```diff
@@ -245,9 +244,52 @@
...
+      "unknownAction": "Leider kann die gewünschte Aktion nicht ausgeführt werden, da sie uns unbekannt ist.",
```
kenodressel, line 250: Could be more user friendly too, and it would be nice to know about. So I suggest adding the help@QH email here.

public/locales/de/translation.json
```diff
@@ -245,9 +244,52 @@
...
+        "changePassword": "Passwort ändern",
+        "success": "Dein Passwort wurde erfolgreich geändert!"
+      },
+      "verifyEmail": {
+        "verificationSuccess": "Deine E-Mail-Adresse wurde erfolgreich verifiziert!",
+        "resendEmail": "E-Mail erneut senden",
+        "emailResent": "Wir haben Dir erneut eine E-Mail mit Verifizierungslink geschickt",
+        "toHome": "Zur Startseite"
+      },
+      "recoverEmail": {
```
kenodressel, line 276: are we even supporting this? cant seem to trigger it anywhere.

src/views/CompleteOfferHelp.jsx
```diff
@@ -48,78 +54,58 @@ export default function CompleteOfferHelp() {
...
-        // We only end up here if there is no email set in the localStorage of the users browser
-        // this might happen e.g. if the user signs up on his desktop pc and clicks the confirmation
-        // link in his mobile phones email client.
-        // eslint-disable-next-line no-alert
-        email = window.prompt(t('views.completeOfferHelp.mailYouRegistered'));
-      }
+    async function completeOfferHelp() {
+      const urlParams = new URLSearchParams(windowLocation.search);
+      const loc = urlParams.get('location');
+      const email = urlParams.get('email');
```
kenodressel, line 60: we should think about a different solution here, this was only intended as a temporary hack since its discouraged by the google firebase docs. If we can't find a better solution though, rock'n'roll :man_dancing:

Maurice22:
You mean sth. like “Please try again. If the error still resides you can contact help@ for support”?

public/locales/de/translation.json
Maurice22: No, we currently don’t. However, I decided to implement it anyway to be future-prove. In my PR description there is a code snippet to test this functionality.

kenodressel:
Yes something like this.

PR #277, opened 2020-05-01
Delete account feature

### LINKED ISSUE #62: Add functionality to remove email addresses (users) from QH
Problem: Firebase is non relational so we need to scan the whole database for these records.

mauriceackel:

* Adds a button to the Dashboard that allows users to delete their account
* Upon clicking, the user has to reenter the PW to confirm the account deletion
* Once the account is deleted in Firebase, a cloud function is triggered, deleting all data associated with the user
* For reported posts, we replace any user did by "ghost-user" in order to not loose this information completely

**TODOs before merge**
- [x] Rename firestore rules on prod /offer-help to /notifications
- [x] Run migration on prod
- [ ] Delete old collection on prod

**Others details**
This feature requires us to rename the collection /offer-help. I renamed it to /notifiactions as this is what is stored in the collection.

Closes #62

### DIFF
--- a/firebase/functions/index.js
+++ b/firebase/functions/index.js
@@ -313,30 +313,31 @@ async function onDeleletedCreate(snap) {
 async function onUserDelete(user) {
   try {
     const db = admin.firestore();
+    const promises = [];
 
     // Delete ask for helps
     const askForHelpEntries = await getEntriesOfUser(db, 'ask-for-help', 'd.uid', user.uid);
-    askForHelpEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'ask-for-help', doc.id));
+    promises.push(askForHelpEntries.docs.map((doc) => deleteDocumentWithSubCollections(db, 'ask-for-help', doc.id)));
 
     // Delete solved posts
     const solvedPostEntries = await getEntriesOfUser(db, 'solved-posts', 'd.uid', user.uid);
-    solvedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'solved-posts', doc.id));
+    promises.push(solvedPostEntries.docs.map((doc) => deleteDocumentWithSubCollections(db, 'solved-posts', doc.id)));
 
     // Delete delted post
     const deletedPostEntries = await getEntriesOfUser(db, 'deleted', 'd.uid', user.uid);
-    deletedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'deleted', doc.id));
+    promises.push(deletedPostEntries.docs.map((doc) => deleteDocumentWithSubCollections(db, 'deleted', doc.id)));
 
     // Delete help offers for all (ask-for-help, solved and deleted)
     const helpOfferEntries = await getEntriesOfUser(db, 'offer-help', 'email', user.email, true);
-    helpOfferEntries.docs.forEach((doc) => doc.ref.update({ email: '', answer: '' }));
+    promises.push(helpOfferEntries.docs.map((doc) => doc.ref.update({ email: '', answer: '' })));
 
     // Delete notifications
     const notificationEntries = await getEntriesOfUser(db, 'notifications', 'd.uid', user.uid);
-    notificationEntries.docs.forEach((doc) => doc.ref.delete());
+    promises.push(notificationEntries.docs.map((doc) => doc.ref.delete()));
 
     // Anonymize reported by
     const reportedPostsEntries = await getEntriesOfUser(db, 'reported-posts', 'uid', user.uid);
-    reportedPostsEntries.docs.forEach((doc) => doc.ref.update({ uid: 'ghost-user' }));
+    promises.push(reportedPostsEntries.docs.map((doc) => doc.ref.update({ uid: 'ghost-user' })));
 
     // Anonymize reported-by
     const reportedEntryIds = reportedPostsEntries.docs.map((doc) => doc.data().askForHelpId);
@@ -350,11 +351,13 @@ async function onUserDelete(user) {
     reportedEntries.forEach((doc) => {
       if (!doc.exists) return;
 
-      doc.ref.update({ 'd.reportedBy': admin.firestore.FieldValue.arrayRemove(user.uid) });
+      promises.push(doc.ref.update({ 'd.reportedBy': admin.firestore.FieldValue.arrayRemove(user.uid) }));
       if (!doc.data().d.reportedBy.includes('ghost-user')) {
-        doc.ref.update({ 'd.reportedBy': admin.firestore.FieldValue.arrayUnion('ghost-user') });
+        promises.push(doc.ref.update({ 'd.reportedBy': admin.firestore.FieldValue.arrayUnion('ghost-user') }));
       }
     });
+
+    await Promise.all(promises);
   } catch (e) {
     // eslint-disable-next-line no-console
     console.error(e);
--- a/firebase/functions/index.js
+++ b/firebase/functions/index.js
@@ -328,7 +328,7 @@ async function onUserDelete(user) {
 
     // Delete help offers for all (ask-for-help, solved and deleted)
     const deletedPostOffersEntries = await getEntriesOfUser(db, 'offer-help', 'email', user.email, true);
-    deletedPostOffersEntries.docs.forEach((doc) => doc.ref.delete());
+    deletedPostOffersEntries.docs.forEach((doc) => doc.ref.update({ email: '', answer: '' }));
 
     // Delete notifications
     const notificationEntries = await getEntriesOfUser(db, 'notifications', 'd.uid', user.uid);
--- a/firebase/functions/index.js
+++ b/firebase/functions/index.js
@@ -338,18 +338,22 @@ async function onUserDelete(user) {
     const reportedPostsEntries = await getEntriesOfUser(db, 'reported-posts', 'uid', user.uid);
     reportedPostsEntries.docs.forEach((doc) => doc.ref.update({ uid: 'ghost-user' }));
 
-    // Anonymize ask-for-help reported-by
+    // Anonymize reported-by
     const reportedEntryIds = reportedPostsEntries.docs.map((doc) => doc.data().askForHelpId);
-    const askForHelpPromise = await db.collection('ask-for-help').where('d.uid', 'in', reportedEntryIds).get();
-    const solvedPromise = await db.collection('solved-posts').where('d.uid', 'in', reportedEntryIds).get();
-    const deletedPromise = await db.collection('deleted').where('d.uid', 'in', reportedEntryIds).get();
+    const entryRefs = [];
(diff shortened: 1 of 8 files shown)


firebase/functions/index.js
```diff
@@ -312,6 +317,58 @@ async function onDeleletedCreate(snap) {
...
+    notificationEntries.docs.forEach((doc) => doc.ref.delete());
+
+    // Anonymize reported by
+    const reportedPostsEntries = await getEntriesOfUser(db, 'reported-posts', 'uid', user.uid);
+    reportedPostsEntries.docs.forEach((doc) => doc.ref.update({ uid: 'ghost-user' }));
+
+    // Anonymize reported-by
+    const reportedEntryIds = reportedPostsEntries.docs.map((doc) => doc.data().askForHelpId);
+    const entryRefs = [];
+    for (let i = 0; i < reportedEntryIds.length; i += 1) {
```
jagaSto, line 351: could do `entryRefs.forEach((entry) => {})` here instead of a primitive for-loop

firebase/functions/index.js
```diff
@@ -312,6 +317,58 @@ async function onDeleletedCreate(snap) {
...
+    }
+
+    const reportedEntries = await db.getAll(...entryRefs);
+    reportedEntries.forEach((doc) => {
+      if (!doc.exists) return;
+
+      doc.ref.update({ 'd.reportedBy': admin.firestore.FieldValue.arrayRemove(user.uid) });
+      if (!doc.data().d.reportedBy.includes('ghost-user')) {
+        doc.ref.update({ 'd.reportedBy': admin.firestore.FieldValue.arrayUnion('ghost-user') });
+      }
```
jagaSto, line 364: should wrap all of these delete operations in a transaction? By doing so, we guarantee that no data will be lost if any of the atomic operations fail (https://firebase.google.com/docs/firestore/manage-data/transactions), and we don't end up in weird intermediate states (e.g. data for the user still exists in some collections, but already was deleted in others)

firebase/functions/utils.js
```diff
@@ -72,8 +72,19 @@ async function deleteDocumentWithSubCollections(db, collectionName, documentId)
...
+async function getEntriesOfUser(db, collection, key, uid, useCollectionGroup = false) {
+  let userEntries;
+  if (!useCollectionGroup) {
+    userEntries = await db.collection(collection).where(key, '==', uid).get();
+  } else {
+    userEntries = await db.collectionGroup(collection).where(key, '==', uid).get();
+  }
+  return userEntries;
+}
+
```
jagaSto, line 84:
```suggestion
  if (!useCollectionGroup) {
    return db.collection(collection).where(key, '==', uid).get();
  } 
  return db.collectionGroup(collection).where(key, '==', uid).get();
}
```

public/locales/de/translation.json
```diff
@@ -248,6 +248,15 @@
     }
   },
   "components": {
+    "deleteAccountButton": {
+      "deleteAccount": "Account löschen",
+      "password": "Passwort",
+      "confirmDeletion": "Löschen bestätigen",
+      "repeatePassword": "Passwort erneut eingeben",
+      "deleteFinally": "Account unwiederuflich löschen",
```
jagaSto, line 256:
```suggestion
      "deleteFinally": "Account unwideruflich löschen",
```

src/views/CompleteNotification.jsx
```diff
@@ -9,15 +9,15 @@ import { isMapsApiEnabled } from '../featureFlags';
...
   const { t } = useTranslation();
   const [isLoading, setLoading] = useState(true);
   const [location, setLocation] = useState('');
   const [error, setError] = useState(undefined);
 
-  const createOfferHelp = async (loc, placeId, email) => {
+  const createNotification = async (loc, placeId, email) => {
     const geofirestore = new GeoFirestore(fb.store);
-    const offerHelpCollection = geofirestore.collection('/offer-help');
+    const notificationCollection = geofirestore.collection('notifications');
```
jagaSto, line 20: Imagine how easy of a change this would have been if we had one central utility for managing firestore interactions 😆

src/views/Dashboard.jsx
```diff
@@ -167,10 +170,79 @@ function Dashboard(props) {
...
+
+    const pw = new FormData(e.target).get('password');
+
+    try {
+      const credentials = await firebase.auth.EmailAuthProvider.credential(user.email, pw);
+      await user.reauthenticateWithCredential(credentials);
+
+      user.delete();
+    } catch (err) {
+      switch (err.code) {
```
jagaSto, line 195: no need to send this error to sentry I assume, right?

Maurice22:
There is a flaw where we don't decrease the response counter in the post so that a post might show more responses as there actually are. This needs to be fixed before merging.

kenodressel:
@Maurice22 not sure if we need to decrease the counter tbh. I think for marketing purposes its fine to leave as is.

Maurice22:
But it leads to a bad UX. Example: A post receives 1 answer. Then, the creator of the answer deletes their account. Then, the creator of the post sees 1 answer, can click the “exand answers” button and will not be displayed anything.
An alternative would be to not delete the answers but replace them with “deleted”.

kenodressel:
Ah okay I thought you were talking about the global couter. Yes it makes sense to decrease the answer counter per entry.

mauriceackel:
@kenodressel @florianschmidt1994 @jagaSto Any opinions?

firebase/functions/index.js
```diff
@@ -305,6 +310,58 @@ async function onDeleletedCreate(snap) {
...
+    // Delete solved posts
+    const solvedPostEntries = await getEntriesOfUser(db, 'solved-posts', 'd.uid', user.uid);
+    solvedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'solved-posts', doc.id));
+
+    // Delete delted post
+    const deletedPostEntries = await getEntriesOfUser(db, 'deleted', 'd.uid', user.uid);
+    deletedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'deleted', doc.id));
+
+    // Delete help offers for all (ask-for-help, solved and deleted)
+    const deletedPostOffersEntries = await getEntriesOfUser(db, 'offer-help', 'email', user.email, true);
```
mauriceackel, line 330:
```suggestion
    const helpOfferEntries = await getEntriesOfUser(db, 'offer-help', 'email', user.email, true);
```

firebase/functions/index.js
```diff
@@ -305,6 +310,58 @@ async function onDeleletedCreate(snap) {
...
+    const solvedPostEntries = await getEntriesOfUser(db, 'solved-posts', 'd.uid', user.uid);
+    solvedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'solved-posts', doc.id));
+
+    // Delete delted post
+    const deletedPostEntries = await getEntriesOfUser(db, 'deleted', 'd.uid', user.uid);
+    deletedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'deleted', doc.id));
+
+    // Delete help offers for all (ask-for-help, solved and deleted)
+    const helpOfferEntries = await getEntriesOfUser(db, 'offer-help', 'email', user.email, true);
+    deletedPostOffersEntries.docs.forEach((doc) => doc.ref.update({ email: '', answer: '' }));
```
mauriceackel, line 331:
```suggestion
    helpOfferEntries.docs.forEach((doc) => doc.ref.update({ email: '', answer: '' }));
```

firebase/functions/index.js
```diff
@@ -305,6 +310,57 @@ async function onDeleletedCreate(snap) {
...
+    const solvedPostEntries = await getEntriesOfUser(db, 'solved-posts', 'd.uid', user.uid);
+    solvedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'solved-posts', doc.id));
+
+    // Delete delted post
+    const deletedPostEntries = await getEntriesOfUser(db, 'deleted', 'd.uid', user.uid);
+    deletedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'deleted', doc.id));
+
+    // Delete help offers for all (ask-for-help, solved and deleted)
+    const helpOfferEntries = await getEntriesOfUser(db, 'offer-help', 'email', user.email, true);
+    helpOfferEntries.docs.forEach((doc) => doc.ref.update({ email: '', answer: '' }));
```
florianschmidt1994, line 331: Could we match via UID here instead of email as well? Don't think that it makes a big difference though

mauriceackel: No, as we don't store a UID here

firebase/functions/index.js
mauriceackel: Not really, due to the way the `deleteDocumentWithSubcollections` is implemented.

firebase/functions/index.js
```diff
@@ -305,6 +310,57 @@ async function onDeleletedCreate(snap) {
...
+
+    // Delete ask for helps
+    const askForHelpEntries = await getEntriesOfUser(db, 'ask-for-help', 'd.uid', user.uid);
+    askForHelpEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'ask-for-help', doc.id));
+
+    // Delete solved posts
+    const solvedPostEntries = await getEntriesOfUser(db, 'solved-posts', 'd.uid', user.uid);
+    solvedPostEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'solved-posts', doc.id));
+
+    // Delete delted post
```
jagaSto, line 325:
```suggestion
    // Delete deleted posts
```

firebase/functions/index.js
```diff
@@ -305,6 +310,57 @@ async function onDeleletedCreate(snap) {
   }
 }
 
+async function onUserDelete(user) {
+  try {
+    const db = admin.firestore();
+
+    // Delete ask for helps
+    const askForHelpEntries = await getEntriesOfUser(db, 'ask-for-help', 'd.uid', user.uid);
+    askForHelpEntries.docs.forEach((doc) => deleteDocumentWithSubCollections(db, 'ask-for-help', doc.id));
```
jagaSto, line 319: `deleteDocumentWithSubCollections` is async, no? Should we not use `await` here/wrap this call with a `Promise.all`?

public/locales/de/translation.json
```diff
@@ -302,6 +302,21 @@
     "statusIndicator": {
       "toHome": "ZUR STARTSEITE"
     },
+    "responses": {
+      "answerDeleted": "Diese Antwort wurde gelöscht.",
+      "answer": "Antworten",
+      "answerFrom": "Antwort von",
```
jagaSto, line 308: I guess these should be plural (`answers`), but not really important

florianschmidt1994:
Let's have a chat tonight before the meeting @kenodressel @mauriceackel to talk about the migration?

PR #278, opened 2020-05-02
Add function to serve hotline numbers from google sheet

florianschmidt1994:
This adds a firebase function `handleIncomingCall` which allows use to expose phone numbers from a call sheet in google drive via http.

```
Twilio <--> Firebase Function <--> Google Sheets API
```

A request to `handleIncomingCall` will return a list of phone numbers that are available to pick up calls right now as comma-separated values. All requests to this endpoint are protected by a token.

**Checklist before deploying this function**
Set the following config values accordingly
- [x] sheet_id
- [x] range
- [x] service_account_email
- [x] private_key
- [x] api_key

Use the command 
```sh
firebase functions:config:set googleSheets.<key>=<VALUE> --project quarantine-hero
```

**Checklist after deploying this function**
- Set the URL of the `handleIncomingCall` in Twilio studio

**Others details**
- [x] This PR introduces a new dependency. Reason:  `googleapis` for access to google sheets via API  

There unfortunately is no issue to track this right now, as I originally wasn't even planning to contribute this to the code base and it evolved from a PoC. Sorry!

### DIFF
--- a/firebase/functions/hotline.js
+++ b/firebase/functions/hotline.js
@@ -1,19 +1,15 @@
 const { google } = require('googleapis');
-const { OAuth2Client } = require('google-auth-library');
-const admin = require('firebase-admin');
 const functions = require('firebase-functions');
 
+const {
+  sheet_id: SPREADSHEET_ID,
+  service_account_email: SERVICE_ACCOUNT_EMAIL,
+  private_key: PRIVATE_KEY,
+  range: RANGE,
+  api_key: API_KEY,
+} = functions.config().googlesheets;
+
 const SCOPES = ['https://www.googleapis.com/auth/spreadsheets'];
-const SPREADSHEET_ID = '1UsyHXG2_XQob7wGCrsSeSGyWt7sgmo-2Ez0ed3JQbdM';
-const RANGE = 'Verfuegbarkeiten!data';
-
-// eslint-disable-next-line camelcase
-const { client_id, client_secret } = functions.config().googlesheets;
-
-const REDIRECT_URL = 'https://europe-west1-qhero-stage.cloudfunctions.net/oauthCallback';
-const DB_TOKEN_PATH = '/google-sheets-api-tokens';
-
-const oauthClient = new OAuth2Client(client_id, client_secret, REDIRECT_URL);
 
 const weekdayIndex = {
   'Monday Morning': 0,
@@ -46,7 +42,7 @@ function getCurrentShift() {
   const now = new Date();
   const day = days[now.getDay()];
 
-  // as long as this function is run in
+  // FIXME: as long as this function is run in
   // 'europe-west1' this will work properly
   // because the time zone is identical to
   // what we assume our call center hours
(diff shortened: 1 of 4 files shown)


firebase/functions/hotline.js
```diff
@@ -0,0 +1,171 @@
...
+  const snapshot = await admin.database().ref(DB_TOKEN_PATH).once('value');
+  const oauthTokens = snapshot.val();
+  oauthClient.setCredentials(oauthTokens);
+
+  return oauthClient;
+}
+
+async function getValues(params) {
+
+  const sheets = google.sheets({ version: 'v4', auth: await getAuthorizedClient() });
```
Maurice22, line 87: Linting issue, should remove padding line

firebase/functions/hotline.js
```diff
@@ -0,0 +1,171 @@
...
+    if (people.length === 0) {
+      // Make it easier for Twilio Flow to check for empty string
+      res.status(200).send('');
+    } else {
+      res.status(200).send(people.join(','));
+    }
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(`An error occurred while retrieving numbers: ${err}`);
+    res.send('').status(500).end();
```
Maurice22, line 163: I think you have to set the status before you call send as it will terminate the response. Furthermore, I think you don't need .end() if you use .send()
Anyways:
```suggestion
    res.status(500).end();
```
Should send the error response without any data.

firebase/functions/hotline.js
```diff
@@ -0,0 +1,171 @@
...
+  return rows
+    .map((row) => {
+      const number = row[1];
+      const availability = row.slice(2, row.length);
+
+      const rowIdx = weekdayIndex[shift];
+
+      return availability[rowIdx] === 'ja' ? number : null;
+    })
+    .filter(Boolean);
```
Maurice22, line 145: What happens here if the number is the empty string?

firebase/functions/hotline.js
```diff
@@ -0,0 +1,171 @@
+const { google } = require('googleapis');
+const { OAuth2Client } = require('google-auth-library');
+const admin = require('firebase-admin');
+const functions = require('firebase-functions');
+
+const SCOPES = ['https://www.googleapis.com/auth/spreadsheets'];
+const SPREADSHEET_ID = '1UsyHXG2_XQob7wGCrsSeSGyWt7sgmo-2Ez0ed3JQbdM';
+const RANGE = 'Verfuegbarkeiten!data';
```
Maurice22, line 8: We should also make those configurable

Maurice22:
Also, I don't quite get how the whole authentication procedure is working. Maybe you could explain why we have to do it this way and why we cannot use a service account for this.

firebase/functions/hotline.js
florianschmidt1994: It will probably fail inside Twilio

Maurice22: I think it will rather be filtered out by the Boolean filter as an empty string is treated as false typically. Anyways, the latter case would actually be the desired one I guess.

florianschmidt1994: You're right. Just tested it

Maurice22:
@florianschmidt1994 Do you prepare prod for this?

PR #281, opened 2020-05-02
Feature/show solved hint from url

jagaSto:

* adds functionality to show the *solved* hint for an entry depending on URL parameters
* adds functionality to show the *deleted* hint for an entry depending on URL parameters
* this enables us to directly link to both features from emails, e.g. when we engage with a help-seeking individual, enabling them to solve their entry directly from the mail, or delete their entry directly from mail. 
* part of the `improve user email engagement` milestone :)

## Preview Solve
Screenshot 2020-05-02 at 22 44 18

## Preview Delete

Screenshot 2020-05-03 at 15 34 41

### DIFF
--- a/src/components/entry/Entry.jsx
+++ b/src/components/entry/Entry.jsx
@@ -33,6 +33,7 @@ export default function Entry(props) {
     reportedBy = [],
     report = false,
     uid = '',
+    showSolveHint = false,
   } = props;
 
   const history = useHistory();
@@ -46,9 +47,9 @@ export default function Entry(props) {
   const [deleted, setDeleted] = useState(false);
   const [solved, setSolved] = useState(false);
   const [attemptingToDelete, setAttemptingToDelete] = useState(false);
-  const [attemptingToSolve, setAttemptingToSolve] = useState(false);
+  const [attemptingToSolve, setAttemptingToSolve] = useState(showSolveHint);
   const [attemptingToReport, setAttemptingToReport] = useState(report);
-  const [popupVisible, setPopupVisible] = useState(false);
+  const [popupVisible, setPopupVisible] = useState(showSolveHint);
 
   const userIsLoggedIn = !!user && !!user.uid;
   const userLoggedInAndReportedEntryBefore = userIsLoggedIn && reportedBy.includes(user.uid);
--- a/src/components/entry/Entry.jsx
+++ b/src/components/entry/Entry.jsx
@@ -34,6 +34,7 @@ export default function Entry(props) {
     report = false,
     uid = '',
     showSolveHint = false,
+    showDeleteHint = false,
   } = props;
 
   const history = useHistory();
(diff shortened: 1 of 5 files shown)


src/util/getUrlParamFromWindowLocation.js
```diff
@@ -0,0 +1,5 @@
+export default function getUrlParamFromWindowLocation(windowLocation, parameter) {
+  // Because we use the hash router, we cannot use the default functionality here as it would expect the query parameters before the hash
+  const urlParams = new URLSearchParams(windowLocation.search);
+  return urlParams.get(parameter);
+}
```
Maurice22, line 5: Can we build this in the same ways useParams from react-router-dom works? So a hook like useQuery() which will return an object with a key-value pair for all query params. I thin this would make it much cleaner as we can call it once and destructure the result

jagaSto: done

Maurice22:
What happens when the user reloads the page? Or navigates away and the back on the site? I assume it will show the dialog again. Should we think about removing the query parameters so that it is kind of a "one-time" thing if you use the special link?

Maurice22:
Fixed conflict

src/util/useQuery.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default () => {
+  const paramObj = {};
+  const paramArray = [...new URLSearchParams(useLocation().search).entries(), ...new URLSearchParams(window.location.search)];
+  paramArray.forEach((entry) => {
+    const [key, value] = entry;
+    paramObj[key] = value;
+  });
+  return paramObj;
+};
```
florianschmidt1994, line 11: This should probably be wrapped in a useEffect() hook in order to avoid rerunning it on every render

src/views/Dashboard.jsx
```diff
@@ -98,6 +101,8 @@ function Dashboard(props) {
             responses={entry.responses}
             reportedBy={entry.reportedBy}
             uid={entry.uid}
+            showSolveHint={attemptingToSolve && !attemptingToDelete && entry.responses > 0 && entryIdFromUrl === entry.id}
```
florianschmidt1994, line 104: Could we move this boolean expression into a variable?

src/components/entry/Entry.jsx
```diff
@@ -45,10 +47,10 @@ export default function Entry(props) {
 
   const [deleted, setDeleted] = useState(false);
   const [solved, setSolved] = useState(false);
-  const [attemptingToDelete, setAttemptingToDelete] = useState(false);
-  const [attemptingToSolve, setAttemptingToSolve] = useState(false);
+  const [attemptingToDelete, setAttemptingToDelete] = useState(showDeleteHint);
+  const [attemptingToSolve, setAttemptingToSolve] = useState(showSolveHint);
```
florianschmidt1994, line 51: Can there be a condition where both are true? I.e. by playing around with query parameters manually? This would probably an edge-case only

Maurice22: This is prevented in the dashboard component

src/util/useQuery.js
Maurice22: Nice idea, I will try this

src/views/Dashboard.jsx
Maurice22: I am not a fan of putting it into a variable as in my opinion it makes it harder to understand what is going on due to a larger distance

PR #298, opened 2020-05-08
Fix issue where token and user object are out of sync

florianschmidt1994:
**The problem**

The user object exposed by firebase has a property called `emailVerified`. This property updates based on the information which is saved in the user database of firebase. It gets updated whenever we load the user information, i.e. via `useAuthState(firebase.auth())`. 
This means that when a user verifies himself and we redirect to another page or a browser refresh happens, that this property is `true`.

Firebase internally uses a JWT for authenticating against its different APIs. This token contains a claim `emailVerified`. This claim is only updated when a user performs a `signIn` or when `user.refreshIdToken(true /* force refresh */)` is called in the code.

It can now happen that a user verifies his email in a different browser than the one he is using to interact with our platform (i.e. verifying on mobile, interacting on desktop).
In this case the user object will be set to `user.emailVerified = true`, whereas the claim will be set to `claims.emailVerified = false`. It can now happen that we allow the user to interact with APIs by showing him some UI based on his `user.emailVerified = true` status, but the subsequent requests to the firebase datasbase will fail with permissions errors because the claims are not properly updated yet. 

**Possible solutions**
 1. Find those edge-cases as described above in our code and call `user.refreshIdToken(true)` in those cases → in my opinion very complex and error prone
2. Use a top-level hook which refreshes the token on each page load → Performs many requests, but relatively safe in terms of consistency
3. Use a hook to check the state and consistency of `emailVerified`.

I've implemented option 3 here because I believe this might be the most robust way to solve this problem. @mauriceackel might also have another concept at hand which we could compare this to. 

The interesting part of this hook is a check where I verify that the claim inside the JWT and the emailVerified property of the user object are in sync. Only if they differ in their state will I perform a forced update, i.e. network request, in order to update the token to match the actual state.

I have yet to fully test this through, but as of now this seems to work in staging.

This code still fails in production because `user.refreshIdToken(true)` is not allowed with our current api key permissions. We (@kenodressel) must add the _Token Service API_ permission to the browser key. I've verified that this makes the difference by adding and removing (and now re-adding) this permission to the staging key. 
Because of this missing permission, currently also the "Bestätigung überprüfen" button does not work in production. (@jagaSto @kenodressel has this ever worked in prod?)

### DIFF
--- a/src/util/emailVerified.js
+++ b/src/util/emailVerified.js
@@ -1,0 +1,49 @@
+import { useEffect, useState } from 'react';
+import { useAuthState } from 'react-firebase-hooks/auth';
+
+export default async function isEmailVerified(user) {
+  if (!user) {
+    return false;
+  }
+
+  const token = await user.getIdTokenResult();
+  const emailVerifiedStateFromToken = token.claims.email_verified;
+
+  if (emailVerifiedStateFromToken !== user.emailVerified) {
+    console.log('Tokens out of sync. Refreshing');
+    const newToken = await user.getIdTokenResult(true);
+    if (newToken.claims.email_verified !== user.emailVerified) {
+      throw new Error('The state of the idToken and user.emailVerified is out of sync even after a refresh');
+    }
+
+    return newToken.claims.emailVerified;
+  }
+
+  return emailVerifiedStateFromToken;
+}
+
+export function useEmailVerified(auth) {
+  const [user, isAuthLoading] = useAuthState(auth);
+
+  const [state, setState] = useState({
+    loading: true,
+    emailVerified: false,
+  });
+
+  useEffect(() => {
+    async function run() {
+      setState({
+        loading: true,
+        emailVerified: false,
+      });
+      const isVerified = await isEmailVerified(user);
+      setState({
+        loading: false,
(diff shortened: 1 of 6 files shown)


kenodressel:
Hard for me to review but reasoning makes sense and code looks good to me.

mauriceackel:
@florianschmidt1994 can you check the build lint and build errors please

src/util/emailVerified.js
```diff
@@ -0,0 +1,49 @@
...
+        emailVerified: false,
+      });
+      const isVerified = await isEmailVerified(user);
+      setState({
+        loading: false,
+        emailVerified: isVerified,
+      });
+    }
+    run();
+  }, [user, isAuthLoading]);
```
mauriceackel, line 46: Have you tested that the hook detects a change in the user object as the verify email property changes? I used something similar elsewhere and it did not work.

florianschmidt1994: We've discussed this in a private call and came to the conclusion, that even if the property change does not trigger an effect call it doesn't matter in our case. We will always execute this hook on page load or when navigating and in this case it will always give us a consistent state

Issue #289, opened 2020-05-06
Find a way to keep up-to-date with sentry error reporting


florianschmidt1994:
It feels like some issues might be going unnoticed and we should find a way to keep up-to-date with this

kenodressel:
We should also check other services:

- [ ] Sendgrid
- [ ] Twilio
- [ ] Firebase

kenodressel:
For Sentry I propose the following:
`Send Alerts for Issues that come up more than 20 times in a single day`

mauriceackel:
@kenodressel Can this be closed?

PR #299, opened 2020-05-10
Tag releases with SHA for sentry

florianschmidt1994:
Add Git SHA to build process and use it to tag sentry releases.

Related to #289

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -125,6 +125,7 @@ const Page = ({ children }) => {
     Sentry.init({
       dsn: process.env.REACT_APP_SENTRY_DSN,
       environment: process.env.REACT_APP_ENV,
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -299,6 +300,21 @@ export default function App() {
...
+              // eslint-disable-next-line no-console
+              console.log('Capturing exception');
+              // eslint-disable-next-line no-console
+              console.log(process.env.REACT_APP_SENTRY_RELEASE);
+              Sentry.captureException('captured test exception');
+            }}
+          >
+            Capture Exception
+          </button>
+        </Route>
```
florianschmidt1994, line 317: I would like to have this on master for testing purposes for a little while, but we can remove it as soon as we believe that everything is working properly

PR #313, opened 2020-05-27
Display new hotline number on landingpage

florianschmidt1994:


florianschmidt1994:
I've also set up the number correctly in Twilio and verified it with a test call

PR #331, opened 2020-06-20
Enhancement/convert firebase functions to ts

jagaSto:

* converts the firebase functions from JS to TS 🚀 
* bumps node version from 8 to 10
* as of now still a draft since there are some open discussion points. 
* the build process needs to be tested on staging and we need to perform some sanity checks that everything is still working!

For references, see:
* https://firebase.google.com/docs/functions/typescript
* https://github.com/firebase/functions-samples/tree/master/typescript-getting-started


firebase/functions/src/types/collections/AskForHelpCollectionEntry.ts
```diff
@@ -0,0 +1,17 @@
+export interface AskForHelpCollectionEntry {
+  uid: string // TODO: check if also available outside of d
```
jagaSto, line 2: can somebody confirm that this is indeed the case? Not visible to me from looking at the staging collection entries

firebase/functions/tslint.json
```diff
@@ -0,0 +1,108 @@
+{
+  "rules": {
```
jagaSto, line 2: default rules created with `firebase init functions`

jagaSto: open for discussion I guess, but I'm pretty much fine with how they are currently!

firebase/functions/src/types/collections/AskForHelpCollectionEntry.ts
kenodressel: its not

firebase/functions/src/types/collections/NotificationsCollectionEntry.ts
```diff
@@ -0,0 +1,13 @@
+export interface NotificationsCollectionEntry {
+  d: {
+    uid: string
+    timestamp: string
```
kenodressel, line 4: should be an integer
```suggestion
    timestamp: number
```

firebase/functions/src/types/collections/AskForHelpCollectionEntry.ts
```diff
@@ -0,0 +1,17 @@
...
+  d: {
+    uid: string
+    timestamp: string
```
kenodressel, line 5: should be int
```suggestion
    timestamp: number
```

firebase/functions/src/types/collections/ReportedPostsCollectionEntry.ts
```diff
@@ -0,0 +1,6 @@
+export interface ReportedPostsCollectionEntry {
+  uid: string
+  timestamp: string
```
kenodressel, line 3:
```suggestion
  timestamp: number
```

firebase/functions/src/types/collections/NotificationsCollectionEntry.ts
```diff
@@ -0,0 +1,13 @@
...
+    email: string
+    location: string
+    plz: string
+    coordinates: FirebaseFirestore.GeoPoint
+    distance?: number // set on searchAndSendNotificationEmails
```
kenodressel, line 9: not a single entry has this property, should we remove it?

jagaSto: as mentioned in the comment, it is set as a part of `searchAndSendNotificationEmails ` to determine the closed offers for an ask-for-help entry, but never persisted in the database

jagaSto:
tested on staging following this list https://github.com/option26/quarantine-hero/issues/335

PR #344, opened 2020-06-27
Feature/Custom implementation of geocoding

mauriceackel:

Added custom geocoding and autocomplete functionality and ditched Google Maps API.

### DIFF
--- a/src/services/GeoService.js
+++ b/src/services/GeoService.js
@@ -4,7 +4,7 @@ const numericRegex = /\d{3,}/;
 const stringRegex = /\D+/;
 
 function enrichEntry(entry) {
-  return { ...entry, description: `${entry.plz || ''}${entry.plz ? ' ' : ''}${entry.name}` };
+  return { ...entry, description: `${entry.plz ? `${entry.plz} ` : ''}${entry.name}` };
 }
 
 function average(numberArray) {
--- a/src/services/GeoService.js
+++ b/src/services/GeoService.js
@@ -49,7 +49,7 @@ export async function getSuggestions(searchString) {
 
 export async function getGeodataForString(searchString) {
   const result = await getSuggestions(searchString);
-  return result[0];
+  return [result];
 }
 
 export async function getGeodataForPlace(placeId) {
--- a/src/services/GeoService.js
+++ b/src/services/GeoService.js
@@ -4,7 +4,12 @@ const numericRegex = /\d{3,}/;
 const stringRegex = /\D+/;
 
 function enrichEntry(entry) {
-  return { ...entry, description: `${entry.plz} ${entry.name}` };
+  return { ...entry, description: `${entry.plz || ''}${entry.plz ? ' ' : ''}${entry.name}` };
+}
+
+function average(numberArray) {
+  const { count, sum } = numberArray.reduce((agg, curr) => ({ count: agg.count + 1, sum: agg.sum + curr }), { count: 0, sum: 0 });
+  return sum / count;
 }
 
 export async function getSuggestions(searchString) {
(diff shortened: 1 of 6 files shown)


src/services/GeoService.js
```diff
@@ -1,55 +1,62 @@
...
-        return reject(status);
-    }
-  }));
+import fb from '../firebase';
+
+const numericRegex = /\d{3,}/;
+const stringRegex = /\D+/;
+
+function enrichEntry(entry) {
+  return { ...entry, description: `${entry.plz || ''}${entry.plz ? ' ' : ''}${entry.name}` };
```
jagaSto, line 7:
```suggestion
  return { ...entry, description: `${entry.plz ? `${entry.plz} ` : ''}${entry.name}` };
```

jagaSto: a bit more readable like this?

src/services/GeoService.js
```diff
@@ -1,55 +1,62 @@
...
-        return resolve(results);
-      }
-      return reject(status);
+export async function getSuggestions(searchString) {
+  const plzMatch = searchString.match(numericRegex);
+  const locationMatch = searchString.match(stringRegex);
+  const location = locationMatch !== null ? locationMatch[0].trim() : '';
+
+  if (plzMatch !== null) {
+    const plz = plzMatch[0];
```
jagaSto, line 21:
```suggestion
    const [ plz ] = plzMatch;
```

src/services/GeoService.js
```diff
@@ -1,55 +1,62 @@
...
-  return new Promise((resolve, reject) => {
-    geocoder.geocode(request, (results, status) => {
-      if (status === 'OK') {
-        return resolve(results);
-      }
-      return reject(status);
+export async function getSuggestions(searchString) {
+  const plzMatch = searchString.match(numericRegex);
+  const locationMatch = searchString.match(stringRegex);
+  const location = locationMatch !== null ? locationMatch[0].trim() : '';
```
jagaSto, line 18: could `locationMatch `ever be an empty array here? If yes, we should check, so `locationMatch[0].trim()` does not break

jagaSto:
found issues on dev so far: 
* filtering with old entries that have a country does not work, e.g. for `http://localhost:3000/#/?address=4020%20Linz,%20%C3%96sterreich`
* input for lower case does not find any matches e.g., try `mannheim` or `berlin`

src/services/GeoService.js
```diff
@@ -1,55 +1,62 @@
...
+        lon: average(aggregatedEntries.map((e) => Number.parseFloat(e.lon))),
+        description: key,
+      };
     });
-  });
+}
+
+export async function getGeodataForString(searchString) {
+  const result = await getSuggestions(searchString);
+  return result[0];
```
jagaSto, line 52:
```suggestion
  return [result];
```

src/services/GeoService.js
```diff
@@ -1,55 +1,62 @@
...
-        return resolve(results);
-      }
-      return reject(status);
+export async function getSuggestions(searchString) {
+  const plzMatch = searchString.match(numericRegex);
+  const locationMatch = searchString.match(stringRegex);
+  const location = locationMatch !== null ? locationMatch[0].trim() : '';
+
+  if (plzMatch !== null) {
+    const [ plz ] = plzMatch;
```
jagaSto, line 21:
```suggestion
    const [plz] = plzMatch;
```

mauriceackel:
WARNING! Fixed case-insensitivity, but this requires to perform a migration on the prod database. Do not merge before!

PR #347, opened 2020-06-27
adds firebase rules to repo and deploy script (fixes #287)

### LINKED ISSUE #287: Publish security rules into repository
Currently blocked by pentesting

kenodressel:

### DIFF
--- a/.github/workflows/deploy-firebase.sh
+++ b/.github/workflows/deploy-firebase.sh
@@ -12,3 +12,4 @@ FIREBASE_PROJECT="quarantine-hero"
 FIREBASE_MESSAGE="QH $GIT_REV deployment to firebase"
 # deploy here because npm does not resolve environment variables in npm run
 npx firebase-tools deploy --only functions --project "$FIREBASE_PROJECT" -m "$FIREBASE_MESSAGE" --non-interactive
+npx firebase-tools deploy --only firestore:rules --project "$FIREBASE_PROJECT" -m "$FIREBASE_MESSAGE" --non-interactive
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-firebase.sh
```diff
@@ -12,3 +12,4 @@ FIREBASE_PROJECT="quarantine-hero"
 FIREBASE_MESSAGE="QH $GIT_REV deployment to firebase"
 # deploy here because npm does not resolve environment variables in npm run
 npx firebase-tools deploy --only functions --project "$FIREBASE_PROJECT" -m "$FIREBASE_MESSAGE" --non-interactive
+npx firebase-tools deploy --only firestore:rules --project "$FIREBASE_PROJECT" -m "$FIREBASE_MESSAGE" --non-interactive
```
mauriceackel, line 15: Is this also deploying the indices?

firebase/firebase.json
```diff
@@ -6,6 +6,11 @@
     "predeploy": [
       "npm --prefix \"$RESOURCE_DIR\" run lint",
       "npm --prefix \"$RESOURCE_DIR\" run build"
-    ]
+    ],
+    "source": "functions"
+  },
```
mauriceackel, line 11: Why is this added?

.github/workflows/deploy-firebase.sh
mauriceackel: Seems that it does not.
```suggestion
npx firebase-tools deploy --only firestore --project "$FIREBASE_PROJECT" -m "$FIREBASE_MESSAGE" --non-interactive
```

firebase/firebase.json
mauriceackel: Looks like functions would be the default here as well so no pain setting it explicitly here guess

PR #361, opened 2020-10-08
SEO improvements v2: update title and noscript content

jagaSto:

* updates `meta title` tag to include `Nachbarschaftshilfe` (as `wir gegen corona` does)
* updates content of `noscript` tag to improve accessibility

To quote from [googles support article](https://support.google.com/webmasters/answer/66353?hl=en#:~:text=Hiding%20text%20or%20links%20in,text%20on%20a%20white%20background):

In general, hidden tex is perceived deceptive, but:

> However, not all hidden text is considered deceptive. For example, if your site includes technologies that search engines have difficulty accessing, like JavaScript, images, or Flash files, using descriptive text for these items can improve the accessibility of your site

> JavaScript: Place the same content from the JavaScript in a <noscript> tag. If you use this method, ensure the contents are exactly the same as what’s contained in the JavaScript, and that this content is shown to visitors who do not have JavaScript enabled in their browser.

Which renders this very basic site if JS is turned off:

Screenshot 2020-10-08 at 22 13 56

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -47,7 +47,7 @@
 </head>
 <body>
   <noscript>
-    <p><b>You need to enable JavaScript to run this app. Otherwise, we show this minified version of the website:</b>
+    <p><b>You need to enable JavaScript to run this app, and to access the full functionality, like asking for help and offering help. Otherwise, we show this minified version of the website:</b>
     </p>
 
     <b>Allgemeines</b>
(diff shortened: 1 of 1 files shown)


kenodressel (approved):
love the solution :rocket:

mauriceackel (approved):
Looks good

PR #369, opened 2020-11-29
Feature/show solved posts on landing page

### LINKED ISSUE #350: Show solved posts on the overview page
list

For solved posts, we do not show the `report entry` flag and put the  `solved` flag there instead. This should make the overview page more engaging and informative!

jagaSto:

* fixes #350 
* would be great if some of you could test this on staging as well to make sure I didn't miss anything :)

_Briefly describe your changes here_

* updates firestore rules to allow read for all solved posts
* adds functionality to show `solved posts` on the landing page, as well as showing them when searching
* to do this, we additionally query the `solved-posts` collection in the `EntryList` (--> solved posts don't appear in the Map view)
* here, we now need to perform client-side sorting in order to merge the results of both collections. Overall, this shouldn't heavily impact the pricing because the number of entries in the `solved post` collection is very marginal. At the same time, once a user marks a post as resolved, the document gets moved from the main collection to the other, so the number of total entities requested does not change significantly.
* *update after discussion* with @mauriceackel and @kenodressel : for now we will only show the solved posts on the landing page to not make it seem empty in case people mark a request as solved. Search and pagination loads will only yield open ask for help entries

_Screenshots_

Desktop:

Screenshot 2020-11-29 at 18 58 22

Mobile:

IMG_0033


src/components/EntryList.jsx
```diff
@@ -82,6 +84,11 @@ export default function EntryList({ pageSize = 0 }) {
     }
   };
 
+  // we cannot use destructuring syntax here because it will break the firebase document entries
+  const getAnnotatedSolvedPosts = (solvedPostsResults) => solvedPostsResults.docs.map((entry) => {
```
jagaSto, line 88: I iniatally wanted to do `solvedPostsResults.docs.map(entry => (...{ entry, solved:true }))` here, but that caused issues with the documents :/

src/components/EntryList.jsx
```diff
@@ -91,17 +98,22 @@ export default function EntryList({ pageSize = 0 }) {
...
-    let documents = results.docs;
+    const solvedPostsResultsAnnotated = getAnnotatedSolvedPosts(solvedPostsResults);
+
+    let documents = [...askForHelpResults.docs, ...solvedPostsResultsAnnotated];
 
     // we need to perform client-side sorting since the location filter is applied
     // https://github.com/quarantine-hero/quarantine-hero/issues/89
     if (searchActive) {
       if (isMapsApiEnabled) {
+        // if maps api is enabled, we retrieve documents from the geo collections, resulting in different structure
```
jagaSto, line 116: can somebody pls double-check me here and make sure that this is correct? I wanted to add a comment why we do the sorting slightly different here (`data().d.timestamp` vs `data().timestamp`)

src/components/EntryList.jsx
mauriceackel: Because your syntax was wrong:
`solvedPostsResults.docs.map(entry => ({ ...entry, solved:true }))`

src/components/EntryList.jsx
mauriceackel: Yeah, you are write, however, this is a bit confusing anyways, because if we would disable the maps api, all new documents would have a different structure than they have when using geofirestore.. Hennyways, the comment is correct

src/components/EntryList.jsx
```diff
@@ -110,33 +122,66 @@ export default function EntryList({ pageSize = 0 }) {
           (doc1, doc2) => doc2.data().d.timestamp - doc1.data().d.timestamp,
         );
       }
+    } else {
+      documents = documents.sort(
+        (doc1, doc2) => doc2.data().d.timestamp - doc1.data().d.timestamp,
+      );
```
mauriceackel, line 128: This basically means that we will always resort on the client, taking the solved posts into account, right?

src/components/EntryList.jsx
jagaSto: yeah sorry, the syntax was of course wrong in the comment^^ nevertheless it doenst work

src/components/EntryList.jsx
jagaSto: yes... the only way to avoid this would be to put all documents in one collection, as we previously discussed. As far as I know, there is no way to directly query from two collections at once

src/components/EntryList.jsx
mauriceackel: Ah sure, the issue are the functions (i.e. .data()) I guess :)

kenodressel:
We should consider halving the page size when building the queries?

src/components/EntryList.jsx
mauriceackel: @kenodressel but we have the maps API currently enabled, so this would defeat the purpose, no? Or do you only mean during search? Even if, I would argue we should keep them as the main reason of this PR is to show that there is activity on the side even if people start flagging their requests as fixed.

mauriceackel:
@kenodressel You mean because we query two collections now?

src/components/EntryList.jsx
kenodressel: Ah we have it enabled because our schema is compatible with the geofirestore? why do we need the flag then?

jagaSto:
@kenodressel @mauriceackel this in now updated, feel free to take another look 🙂  Solved posts only visible on landing page now, also updated the descriptionon

src/components/EntryList.jsx
```diff
@@ -1,176 +1,93 @@
...
+      return;
     }
 
-    return query;
-  };
-
-  const buildFilteredQuery = async (searchAttr, placeId = undefined) => {
-    // Fallback
-    if (!searchAttr) return buildQuery();
+    const oldestEntry = documents.filter((d) => !d.solved).reduce((oldest, doc) => (doc.timestamp < oldest.timestamp ? doc : oldest), documents[0]);
```
jagaSto, line 36:
```suggestion
    const oldestEntry = documents
    .filter((d) => !d.solved)
    .reduce((oldest, doc) => (doc.timestamp < oldest.timestamp ? doc : oldest), documents[0]);
```

 Makes it easier to read. Also, could we add a comment about why we filter the solved posts in the first place? To make sure we still  understand this in 6 months :)

src/components/EntryList.jsx
```diff
@@ -1,176 +1,93 @@
...
 
-
-  const appendDocuments = (documents) => {
-    setLastEntry(documents[documents.length - 1]);
-    const newEntries = documents
-      .map(parseDoc)
-      .filter(Boolean); // filter entries that we weren't able to parse and are therefore null
-    setEntries((e) => [...e, ...newEntries]);
+  const loadMore = async () => {
+    const documents = await loadMoreDocuments(lastEntry.timestamp, pageSize);
```
jagaSto, line 47: `lastEntry` can be undefined here, right?

src/services/loadData.js
```diff
@@ -0,0 +1,57 @@
...
+
+export async function getInitialDocuments(pageSize = 20) {
+  const askForHelpQuery = askForHelpCollection.orderBy('d.timestamp', 'desc').limit(pageSize);
+  const solvedPostsQuery = solvedPostsCollection.orderBy('d.timestamp', 'desc').limit(pageSize);
+
+  const askForHelpDocuments = await askForHelpQuery.get().then((snap) => snap.docs.map(parseDoc).filter(Boolean));
+  const solvedPostsDocuments = await solvedPostsQuery.get().then((snap) => snap.docs.map(markSolved).map(parseDoc).filter(Boolean));
+
+  const oldestAskForHelpTimestamp = Math.min(...askForHelpDocuments.map((d) => d.timestamp));
+  const youngerSolvedPostsDocuments = solvedPostsDocuments.filter((d) => d.timestamp >= oldestAskForHelpTimestamp);
```
jagaSto, line 24: also, a comment here why need do this would be nice for later us :)

src/components/EntryList.jsx
```diff
@@ -1,176 +1,93 @@
...
-    // https://github.com/quarantine-hero/quarantine-hero/issues/89
-    if (searchActive) {
-      if (isMapsApiEnabled) {
-        documents = documents.sort(
-          (doc1, doc2) => doc2.data().timestamp - doc1.data().timestamp,
-        );
+  useEffect(() => {
+    async function init() {
+      if (addressFromUrl) {
+        const geoData = await getGeodataForString(addressFromUrl);
```
jagaSto, line 54: should catch if `geoData` is undefined

Screenshot 2021-01-03 at 13 28 44

src/components/EntryList.jsx
mauriceackel: Well, this shouldn't really happen in practice but in theory it could. Added an early return if lastEntry is undefined

src/components/EntryList.jsx
```diff
@@ -1,176 +1,103 @@
...
+    if (documents.length === 0) {
+      if (clear) {
+        setEntries([]);
+      }
+      return;
     }
 
-    return query;
-  };
+    // We need to filter oout the solved posts here as we want to load the next documents based on the timestamp of the last ask-for-help document
```
jagaSto, line 36:
```suggestion
    // We need to filter out the solved posts here as we want to load the next documents based on the timestamp of the last ask-for-help document
```

src/components/EntryList.jsx
```diff
@@ -1,176 +1,103 @@
...
-    if (addressFromUrl) {
-      setLocation(addressFromUrl);
-      loadDocuments(buildFilteredQuery(addressFromUrl), true);
-    } else {
-      setLocation('');
-      loadDocuments(buildQuery());
+    async function init() {
+      if (addressFromUrl) {
+        const geoData = await getGeodataForString(addressFromUrl);
+        if (geoData === undefined) {
```
jagaSto, line 62: is this case logging worthy?

src/services/loadData.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export async function getInitialDocuments(pageSize = 20) {
+  const askForHelpQuery = askForHelpCollection.orderBy('d.timestamp', 'desc').limit(pageSize);
+  const solvedPostsQuery = solvedPostsCollection.orderBy('d.timestamp', 'desc').limit(pageSize);
+
+  const askForHelpDocuments = await askForHelpQuery.get().then((snap) => snap.docs.map(parseDoc).filter(Boolean));
+  const solvedPostsDocuments = await solvedPostsQuery.get().then((snap) => snap.docs.map(markSolved).map(parseDoc).filter(Boolean));
+
+  // We want to filter out all solved posts that are older than the last ask-for-help we loaded
```
jagaSto, line 23: one possible disadvantage of this that just came to my mind: consider this scenario for our history of latest entries on the landing page:
```
ask-for-help-entry-date | status
today | solved
today | **open**
today | solved
today | solved
today | solved
today | solved
today | solved
today | solved
today | solved
today | solved
today | solved
```

--> based on this logic, we would omit all records past record two, right? Or even, better, if only the latest entry would be not solved yet, we would not show any of the other solved ones

src/components/EntryList.jsx
mauriceackel: Yeah, we could add a Sentry logging here. It will effectively fire whenever a user klicks on the location of an entry and this location is undefined.

src/services/loadData.js
mauriceackel: No that's not how it works, as we query X 'open' entries anyways.

jagaSto: f2f: discussed -->this scenario wont happen because we will always load at least X ask-for-help entries (in this, case, X = 20)

src/components/EntryList.jsx
```diff
@@ -1,176 +1,103 @@
...
-        documents = documents.sort(
-          (doc1, doc2) => doc2.data().timestamp - doc1.data().timestamp,
-        );
-      } else {
-        documents = documents.sort(
-          (doc1, doc2) => doc2.data().d.timestamp - doc1.data().d.timestamp,
-        );
-      }
+  const loadMore = async () => {
+    if (lastEntry === undefined) {
```
kenodressel, line 51: maybe a general note:
- I would rather use `typeof lastEntry === 'undefined'`, this ensure that even in the case that the variable is never defined, it does not throw a reference error: https://stackoverflow.com/questions/4725603/variable-undefined-vs-typeof-variable-undefined

src/services/loadData.js
```diff
@@ -0,0 +1,59 @@
...
+
+  return combinedDocs;
+}
+
+export async function loadMoreDocuments(startAfter, pageSize = 20) {
+  const query = askForHelpCollection.orderBy('d.timestamp', 'desc').startAfter(startAfter).limit(pageSize);
+
+  const snapshot = await query.get();
+  // Filter broken documents
+  const documents = snapshot.docs.map(parseDoc).filter(Boolean);
```
kenodressel, line 40: what are these broken documents?

src/views/AskForHelp.jsx
```diff
@@ -24,22 +23,14 @@ export default function AskForHelp() {
...
-        throw new Error('PlaceId was undefined in ask-for-help');
-      }
-      const result = await getGeodataForPlace(placeId);
-
-      plz = result.plz;
-      const coordinates = getLatLng(result);
-      lat = coordinates.lat;
-      lng = coordinates.lng;
+    if (!placeId) {
+      throw new Error('PlaceId was undefined in ask-for-help');
```
kenodressel, line 27: shouldn't we handle this?, eg show the user an error message?

kenodressel (approved):

src/services/loadData.js
mauriceackel: I have no idea 😄 I think this filter was implemented by @florianschmidt1994 once and I rust continued using it

src/views/AskForHelp.jsx
mauriceackel: Good question.. However, I would create a separat issue for this as it is already an issue in the current code and not specific to this refactoring

src/components/EntryList.jsx
mauriceackel: Eslint is preventing us from using variables that have not been set, I guess.. So I think it is fine

src/components/EntryList.jsx
mauriceackel: Done

mauriceackel (approved):

PR #373, opened 2020-12-05
Added donate button to share buttons

mauriceackel:

This PR adds a donation button next to our social buttons. The links behind it leads to a PayPal donation campaign created with the Option 26 e.V. organization account.


kenodressel:
Can we continue this now?

mauriceackel:
I think it depends on whether or not we want to have a public donate button or not. Technically, it should work just fine and organization-wise, everything should be sorted out soon.

PR #377, opened 2021-01-04
Improved wording for PW reset and fixed layout

### LINKED ISSUE #375: easier unsubscription
We are getting some emails from users who are unsure how they can ubnsubscribe/delete their account. Perhaps this is caused by the fact that they can't login to the dashboard as they never had to set a password when registering for notifications.

Suggested workflow: Our emails already have an "unsubscribe" button I guess but I don't know what it does. Perhaps, we want to change this to redirect the user to the dashboard with a link that already includes the email. When opening the dashboard with this additional URL param, we could check whether the passed email is registered but does not have a PW set yet (not sure how we can do that). If there is no PW set yet, we would render a different page that shows only the option to get an email with a link to set the PW.

### LINKED ISSUE #381: Improve wording for post requests

mauriceackel:

This PR arose from #375. However, it seems to be unfeasible what is described there. As a result, I only improved the wording for the PW reset button to signal, that users who haven't set their password yet (i.e. just enabled notification) can set a password using this button too.

This also handles the wording change of the placeholder when posting a request.

Will somewhat close #375.
Closes #381 

Bildschirmfoto 2021-01-07 um 11 30 57
Bildschirmfoto 2021-01-07 um 11 31 30
Bildschirmfoto 2021-01-08 um 12 47 19

### DIFF
--- a/public/locales/de/translation.json
+++ b/public/locales/de/translation.json
@@ -86,6 +86,7 @@
       "thisFunction": "diese Funktion",
       "whereAreYou": "Wo bist Du?",
       "whatCanWeDo": "Wobei kann man Dir helfen?",
+      "describeRequest": "Beschriebe hier, bei was Du Hilfe benötigst. Dieser Text wird potentiellen Helfenden angezeigt.",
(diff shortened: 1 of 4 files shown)


jagaSto (approved):
lgtm

PR #378, opened 2021-01-04
Fixing undefined email action error

mauriceackel:

Our currently most triggered error occurs in HandelEmailActions.jsx, whenever the mode of the firebase action is undefined. This currently happens a lot as the custom `useQuery()` hook returns an empty object at first, which makes `mode` undefined and hence leads to the error.

This fix is a rewrite of the `useQuery()` hook so that it does not return an empty object initially. The code is inspired by this https://github.com/ReactTraining/react-router/blob/085d0a862c33f9cfbbf825fddb643cbc0ed44243/packages/react-router-dom/examples/QueryParams.js#L29 and the code we already had in place

### DIFF
--- a/src/util/useQuery.js
+++ b/src/util/useQuery.js
@@ -5,7 +5,7 @@ export default () => {
   const { search } = useLocation();
 
   return React.useMemo(() => {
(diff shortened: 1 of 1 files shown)


jagaSto (approved):
tested signup and PW reset, LGTM

PR #388, opened 2021-02-24
Feature/hotline

mauriceackel:

This PR introduces a Twilio Serverless project to the code, which handles the new hotline functionality. It also takes over the functionality of retrieving the list of people who are currently on the hotline shift. Hence, this functionality is removed from the firebase backend.


kenodressel:
Why do the files have `.protected` in their name?

mauriceackel:
This is due to the way twilio handles accessibility in their Serverless infrastructure:
https://www.twilio.com/docs/labs/serverless-toolkit/general-usage
https://www.twilio.com/docs/runtime/functions-assets-api/api/understanding-visibility-public-private-and-protected-functions-and-assets

Adding .protected means that the files are only accessible from within twilio and not by the public.

PR #392, opened 2021-03-10
feat: add max amount of open requests per user

jagaSto:

* adds a feature to limit the number of open requests a user can have to avoid potential spam.
* in this case, we show a short hint explaining to the user that they should close one of their open entries first.
* hotline account can always create new posts

## Preview

Screenshot 2021-03-10 at 21 33 26

### DIFF
--- a/src/views/AskForHelp.jsx
+++ b/src/views/AskForHelp.jsx
@@ -1,13 +1,14 @@
-import React, { useState } from 'react';
+import React, { useState, useMemo } from 'react';
 import { GeoFirestore } from 'geofirestore';
 import { Redirect, useHistory, Link } from 'react-router-dom';
+import { useCollectionData } from 'react-firebase-hooks/firestore';
 import { useAuthState } from 'react-firebase-hooks/auth';
 import { useTranslation } from 'react-i18next';
 import fb from '../firebase';
 import LocationInput from '../components/LocationInput';
 import { getGeodataForPlace, getLatLng } from '../services/GeoService';
 import { useEmailVerified } from '../util/emailVerified';
-import { hotlineId } from '../appConfig';
+import { hotlineId, allowedOpenRequestsPerUser } from '../appConfig';
 
 export default function AskForHelp() {
   const { t } = useTranslation();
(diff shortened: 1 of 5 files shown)


src/views/AskForHelp.jsx
```diff
@@ -24,6 +25,15 @@ export default function AskForHelp() {
 
   const isHotline = user && user.uid === hotlineId;
```
mauriceackel, line 26:
```suggestion
  const isHotline = useMemo(() => user && user.uid === hotlineId, [user, hotlineId]);
```

If we include this constant in the dependency array below, we should memorize it, otherwise the useMemo hook below will be recomputed unnecessarily

src/views/AskForHelp.jsx
```diff
@@ -24,6 +25,15 @@ export default function AskForHelp() {
...
 
+  const askForHelpCollection = fb.store.collection('ask-for-help');
+
+  const [existingDocsForUser] = useCollectionData(
+    user ? askForHelpCollection.where('d.uid', '==', user.uid) : null,
+    { idField: 'id' },
+  );
+
+  const showAllowedRequestsHint = useMemo(() => !isHotline || (existingDocsForUser && existingDocsForUser.length > allowedOpenRequestsPerUser), [existingDocsForUser]);
```
mauriceackel, line 35:
```suggestion
  const showAllowedRequestsHint = useMemo(() => !isHotline || (existingDocsForUser && existingDocsForUser.length > allowedOpenRequestsPerUser), [existingDocsForUser, isHotline]);
```

We have to add `isHotline` in the dependency array here.

mauriceackel, line 35: Wouldn't we need to test for `existingDocsForUser.length >= allowedOpenRequestsPerUser`, otherwise we would allow 4 instead of 3 requests?

mauriceackel (changes requested):

src/views/AskForHelp.jsx
jagaSto: yeah lol 😄

mauriceackel (approved):

PR #396, opened 2021-03-14
fix: change font weight to font-normal

jagaSto:
https://user-images.githubusercontent.com/20641332/111076793-21422d00-84ee-11eb-9b22-62a79053de29.mov


mauriceackel:
Not relevant anymore

PR #400, opened 2021-03-25
Slackops/message on solve

jagaSto:

* this PR posts a reply into the slack thread if a user marks the entry as solved


firebase/functions/src/types/interface/collections/SolvedPostsCollectionEntry.ts
```diff
@@ -1,4 +1,3 @@
 import { AskForHelpCollectionEntry } from './AskForHelpCollectionEntry';
 
-// tslint:disable-next-line:no-empty-interface
-export interface SolvedPostsCollectionEntry extends AskForHelpCollectionEntry { }
+export type SolvedPostsCollectionEntry = AskForHelpCollectionEntry;
```
jagaSto, line 3: we need to use a type alias here since extending the previous interface with an empty interface leads to an empty interface^^

mauriceackel (approved):
LGTM

